PR #3, opened 2020-04-28
Add upgrade apis

wanghaoran1988:


deploy/crds/upgrade.managed.openshift.io_upgradeconfigs_crd.yaml
```diff
@@ -0,0 +1,225 @@
...
+            and upgrade window and freeze window
+          properties:
+            desired:
+              description: Specify the desired release
+              properties:
+                channel:
+                  description: Channel we gonna use for upgrades
+                  type: string
+                force:
+                  default: false
```
mrbarge, line 48: When testing, I couldn't create the CRD due to this:
`The CustomResourceDefinition "upgradeconfigs.upgrade.managed.openshift.io" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[desired].properties[force].de
fault: Forbidden: must not be set`

I tried on a minikube (1.18) and an OSD cluster. After a bit of research I [noticed a comment elsewhere](https://github.com/coreos/prometheus-operator/issues/3122#issuecomment-610419546) noting that defaults are only supported in CRDs using apiextensions v1.

mrbarge, line 48: When testing, I couldn't create the CRD due to this:
`The CustomResourceDefinition "upgradeconfigs.upgrade.managed.openshift.io" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[desired].properties[force].default: Forbidden: must not be set`

I tried on a minikube (1.18) and an OSD cluster. After a bit of research I noticed a comment elsewhere noting that defaults are only supported in CRDs using apiextensions v1.

wanghaoran1988: Yes, the crds are generated from v1beta1(which is default), I will regenerate using v1

deploy/crds/upgrade.managed.openshift.io_upgradeconfigs_crd.yaml
```diff
@@ -0,0 +1,224 @@
...
+              and upgrade window and freeze window
+            properties:
+              desired:
+                description: Specify the desired release
+                properties:
+                  channel:
+                    description: Channel we gonna use for upgrades
+                    type: string
+                  force:
+                    default: false
```
bmeng, line 48: Is the default value redundant if it is required?

deploy/crds/upgrade.managed.openshift.io_upgradeconfigs_crd.yaml
```diff
@@ -0,0 +1,224 @@
...
+                  maximumCount:
+                    description: Maximum customerFreezeWindow can be created
+                    format: int32
+                    minimum: 0
+                    type: integer
+                  maximumDuration:
+                    description: Maximum duration for a customer freeze window
+                    properties:
+                      unitOfMeasure:
+                        description: 'Valid values are: "Day", "week", "Month"'
```
bmeng, line 71: week should be Week?

pkg/apis/upgrade/v1alpha1/upgradeconfig_types.go
```diff
@@ -0,0 +1,164 @@
...
+	StartTime *metav1.Time `json:"startTime,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	CompleteTime *metav1.Time `json:"completeTime,omitempty"`
+}
+type UpgradeConditionType string
+
+const (
+	UpgradeValidated UpgradeConditionType = "Validated"
+	UpgradeComplete  UpgradeConditionType = "Completed"
```
bmeng, line 34: UpgradeCompleted

pkg/apis/upgrade/v1alpha1/upgradeconfig_types.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+type UpgradePhase string
+
+const (
+	UpgradePhaseNew       UpgradePhase = "New"
+	UpgradePhasePending   UpgradePhase = "Pending"
+	UpgradePhaseUpgrading UpgradePhase = "Upgrading"
+	UpgradePhaseUpgraded  UpgradePhase = "Upgraded"
+	UpgradePhaseFaild     UpgradePhase = "Failed"
```
bmeng, line 64: UpgradePhaseFailed

deploy/crds/upgrade.managed.openshift.io_upgradeconfigs_crd.yaml
wanghaoran1988: I don't think so

deploy/crds/upgrade.managed.openshift.io_upgradeconfigs_crd.yaml
wanghaoran1988: Good catch

pkg/apis/upgrade/v1alpha1/upgradeconfig_types.go
wanghaoran1988: I prefer not have the "Upgrade" prefix, We know it's talking about upgrade even don't have it, right?

PR #6, opened 2020-05-12
Add test framework

sedroche:
https://issues.redhat.com/browse/OSD-3622

Can be run with `go test ./...`


sedroche:
@wanghaoran1988 @mrbarge Does this look ok? (Pretty much replicating the gcp-operator testing setup)
I wasn't sure about `pkg/util/mocks/cr-client.go ` which seems to be a `mockgen` mock of `sigs.k8s.io/controller-runtime@v0.5.2/pkg/client/interfaces.go` to work with `GoMock`

If this all looks good i'll add in some docs etc

sedroche:
@superfunhappytime https://github.com/openshift/managed-upgrade-operator/pull/6/commits/a95fcf163a868a853b6d147c9e9fc3e7593947f9

mrbarge:
@sedroche - looks great. 

Regarding the mock of the controller-runtime client, I see there is also a fake client that could be used as well: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/fake/client.go

Looks like the cloud-ingress-operator is using it. To be honest I have used neither in any degree of seriousness to vouch for one or the other, but we do get the fakeClient  without needing to mockgen.

sedroche:
Thanks Matt. I have used the sigs fake client before and I personally would prefer to make use of that - ready made - client rather than creating a separate mock.
I can only presume the gcp operator did this for consistency in testing both clients [it has a separate gcp client as well](https://github.com/openshift/gcp-project-operator/blob/master/pkg/gcpclient/client.go))

I don't foresee managed-upgrade-operator needing another client.

Example with BDD fakeclient looks like 
```go
var _ = Describe("UpgradeConfigController", func() {
	var (
		reconciler reconcile.Reconciler
		testScheme *runtime.Scheme
	)

	BeforeEach(func() {
		testScheme = runtime.NewScheme()
		_ = upgradev1alpha1.SchemeBuilder.AddToScheme(testScheme)
	})
	Context("Reconcile", func() {
		var upgradeConfigName types.NamespacedName

		BeforeEach(func() {
			upgradeConfigName = types.NamespacedName{
				Name:      "test-upgradeconfig",
				Namespace: "test-namespace",
			}

			reconciler, _ = NewReconcileUpgradeConfig(
				fake.NewFakeClientWithScheme(testScheme, &upgradev1alpha1.UpgradeConfig{
					ObjectMeta: metav1.ObjectMeta{
						Name:      upgradeConfigName.Name,
						Namespace: upgradeConfigName.Namespace,
					},
				}),
				testScheme,
			)
		})
		It("Returns without error", func() {
			_, err := reconciler.Reconcile(reconcile.Request{NamespacedName: upgradeConfigName})
			Expect(err).NotTo(HaveOccurred())
		})
	})
})
```

wdyt?

mrbarge:
Yeah that looks great! I would be all in favour of switching to it personally.

sedroche:
@mrbarge Some initial docs and make command have ben added. If you are good ill merge

standard.mk
```diff
@@ -0,0 +1,3 @@
+.PHONY: gotest
+gotest:
```
wanghaoran1988, line 2: rename to "test" ?

sedroche: Yeah, makes it easier, ill update :+1:

PR #7, opened 2020-05-13
freezewindowvalidation tmpl - OSD-3515

himanshudogra:


sedroche:
Template lgtm but are these webhooks going into https://github.com/openshift/managed-cluster-validating-webhooks? @mrbarge @dofinn ?

wanghaoran1988:
@sedroche Yes, it should goes into https://github.com/openshift/managed-cluster-validating-webhooks

PR #9, opened 2020-05-28
Add upgrade controller

wanghaoran1988:


sedroche:
:eyes:

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -0,0 +1,768 @@
...
+			return false, err
+		}
+	}
+
+	return true, nil
+}
+
+// This will update the 3rd subscriptions
+func UpdateSubscriptions(c client.Client, upgradeConfig *upgradev1alpha1.UpgradeConfig, reqLogger logr.Logger) (bool, error) {
+	for _, item := range upgradeConfig.Spec.SubscriptionUpdates {
```
sedroche, line 339: Does this need to be validated too?

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -0,0 +1,768 @@
...
+			condition.StartTime = &metav1.Time{time.Now()}
+			conditions.SetCondition(*condition)
+			history.Conditions = conditions
+			upgradeConfig.Status.History.SetHistory(*history)
+			err := cl.Status().Update(context.TODO(), upgradeConfig)
+			if err != nil {
+				return err
+			}
+		}
+		if condition.Status == corev1.ConditionTrue {
```
sedroche, line 489: We should reconcile on each stage every (state should be based on the cluster state not our status) so I would remove this

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -0,0 +1,768 @@
...
+}
+
+// This trigger the upgrade process
+func (c ClusterUpgrader) UpgradeCluster(cl client.Client, upgradeConfig *upgradev1alpha1.UpgradeConfig, reqLogger logr.Logger) error {
+
+	log.Info("upgrading cluster")
+	history := upgradeConfig.Status.History.GetHistory(upgradeConfig.Spec.Desired.Version)
+	conditions := history.Conditions
+
+	history.Phase = upgradev1alpha1.UpgradePhaseUpgrading
```
sedroche, line 465: For correctness should this be set within `CommenceUpgrade` func?

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -0,0 +1,768 @@
...
+// Check whether nodes are upgraded or not
+func NodesUpgraded(c client.Client, nodeType string, reqLogger logr.Logger) (bool, error) {
+	configPool := &machineconfigapi.MachineConfigPool{}
+	err := c.Get(context.TODO(), types.NamespacedName{Name: nodeType}, configPool)
+	if err != nil {
+		return false, nil
+	}
+	//TODO send timeout alert if wait timeout
+	if configPool.Status.ReadyMachineCount != configPool.Status.UpdatedMachineCount {
+		errMsg := fmt.Sprintf("not all %s are upgraded, upgraded: %v, ready: %v", nodeType, configPool.Status.UpdatedMachineCount, configPool.Status.ReadyMachineCount)
```
mrbarge, line 432: In the existing upgrade script, it compares the total number of machines in the pool (regardless of their "ready" state) vs the number updated, in order to determine if all workers are upgraded.
I mention it because I'm seeing some odd behaviour when running an upgrade with the operator in its current state. It seems to not wait at this step at all (eg I never see the "not all X are upgraded" anywhere in the logs). Not entirely sure that this check is the problem, but something doesn't look right.

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -0,0 +1,768 @@
...
+			upgradev1alpha1.UpgradeValidated:              ValidateUpgradeConfig,
+			upgradev1alpha1.UpgradePreHealthCheck:         PreClusterHealthCheck,
+			upgradev1alpha1.UpgradeScaleUpExtraNodes:      EnsureExtraUpgradeWorkers,
+			upgradev1alpha1.ControlPlaneMaintWindow:       CreateControlPlaneMaintWindow,
+			upgradev1alpha1.CommenceUpgrade:               CommenceUpgrade,
+			upgradev1alpha1.ControlPlaneUpgraded:          ControlPlaneUpgraded,
+			upgradev1alpha1.AllMastersUpgraded:            AllMastersUpgraded,
+			upgradev1alpha1.RemoveControlPlaneMaintWindow: RemoveControlPlaneMaintWindow,
+			upgradev1alpha1.WorkersMaintWindow:            CreateWorkerMaintWindow,
+			upgradev1alpha1.AllWorkerUpgraded:             AllWorkersUpgraded,
```
mrbarge, line 82: Recommend we rename this to "AllWorkersUpgraded" (add "s" in Workers), so that it is consistent with the function name.

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -0,0 +1,768 @@
...
+			upgradev1alpha1.RemoveMaintWindow:             RemoveMaintWindow,
+			upgradev1alpha1.PostClusterHealthCheck:        PostClusterHealthCheck,
+		}
+	})
+
+	return upgrader
+}
+
+var postFixes = map[string]func(c client.Client, upgradeConfig *upgradev1alpha1.UpgradeConfig, reqLogger logr.Logger) (bool, error){
+	"4.3.0": func(c client.Client, upgradeConfig *upgradev1alpha1.UpgradeConfig, reqLogger logr.Logger) (bool, error) {
```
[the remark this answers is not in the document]
sedroche: Also I don't think this should be committed as 4.3 is completed

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -0,0 +1,768 @@
...
+
+// Check whether nodes are upgraded or not
+func NodesUpgraded(c client.Client, nodeType string, reqLogger logr.Logger) (bool, error) {
+	configPool := &machineconfigapi.MachineConfigPool{}
+	err := c.Get(context.TODO(), types.NamespacedName{Name: nodeType}, configPool)
+	if err != nil {
+		return false, nil
+	}
+	//TODO send timeout alert if wait timeout
+	if configPool.Status.ReadyMachineCount != configPool.Status.UpdatedMachineCount {
```
mrbarge, line 431: This needs to compare against the total number of machines, not the number of ready machines. (the number of ready machines seems to always equal the number of upgraded machines). It will match the behaviour of the original script with that change. I've tested the suggestion below and now it properly waits for nodes to be upgraded before progressing to the further steps.

```suggestion
	if configPool.Status.MachineCount != configPool.Status.UpdatedMachineCount {
```

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -0,0 +1,768 @@
...
+	log.Info("compare two versions")
+	sort.Strings(versions)
+	if versions[0] != current {
+		log.Info(fmt.Sprintf("validation failed, current version %s is greater than desired %s", current, upgradeConfig.Spec.Desired.Version))
+
+		return false, fmt.Errorf("desired version %s is greater than current version %s", upgradeConfig.Spec.Desired.Version, current)
+	}
+
+	// Update the channel if it's not match
+	if clusterVersion.Spec.Channel != upgradeConfig.Spec.Desired.Channel {
```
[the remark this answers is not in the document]
wanghaoran1988: For some reason we may need update the channel I think as different channel has different edges

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -0,0 +1,768 @@
...
+			}
+			return false, nil
+		})
+	}
+
+	// Check whether the desired version exists in availableUpdates
+	avalibleVersions := clusterVersion.Status.AvailableUpdates
+	found := false
+	for _, v := range avalibleVersions {
+		if v.Version == upgradeConfig.Spec.Desired.Version {
```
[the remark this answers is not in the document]
wanghaoran1988: good catch, let's add that to align with current scripts for now

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -0,0 +1,768 @@
...
+}
+
+// ClusterHealthCheck performs cluster healthy check
+func PreClusterHealthCheck(c client.Client, upgradeConfig *upgradev1alpha1.UpgradeConfig, reqLogger logr.Logger) (bool, error) {
+
+	return performClusterHealthCheck(c)
+}
+
+// This will create a new machineset with 1 extra replicas for workers in every region
+func EnsureExtraUpgradeWorkers(c client.Client, upgradeConfig *upgradev1alpha1.UpgradeConfig, reqLogger logr.Logger) (bool, error) {
```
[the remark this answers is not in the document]
wanghaoran1988: I am not sure about this, as we have so many upgrade check steps,  `postUpgradeVerification` will check daemonsets/replicasets status after upgrade, I don't think we should watch all of them ?

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -0,0 +1,768 @@
...
+
+func NewUpgrader(reqLogger logr.Logger, client client.Client) ClusterUpgrader {
+	once.Do(func() {
+		upgrader = map[upgradev1alpha1.UpgradeConditionType]UpgradeStep{
+			upgradev1alpha1.UpgradeValidated:              ValidateUpgradeConfig,
+			upgradev1alpha1.UpgradePreHealthCheck:         PreClusterHealthCheck,
+			upgradev1alpha1.UpgradeScaleUpExtraNodes:      EnsureExtraUpgradeWorkers,
+			upgradev1alpha1.ControlPlaneMaintWindow:       CreateControlPlaneMaintWindow,
+			upgradev1alpha1.CommenceUpgrade:               CommenceUpgrade,
+			upgradev1alpha1.ControlPlaneUpgraded:          ControlPlaneUpgraded,
```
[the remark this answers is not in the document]
wanghaoran1988: control plane is for checking all the CO is upgraded, and the current version in `clusterversion` changed, `AllMastersUpgraded` is checking whether the master workers are upgrade or not.

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -0,0 +1,768 @@
...
+		if v.Version == upgradeConfig.Spec.Desired.Version {
+			found = true
+		}
+	}
+
+	if !found {
+		log.Info(fmt.Sprintf("failed to find the desired version %s in channel %s", upgradeConfig.Spec.Desired.Version, upgradeConfig.Spec.Desired.Channel))
+		//We need update the condition
+		errMsg := fmt.Sprintf("cannot find version %s in available updates", upgradeConfig.Spec.Desired.Version)
+		err = c.Get(context.TODO(), types.NamespacedName{Name: upgradeConfig.Name}, upgradeConfig)
```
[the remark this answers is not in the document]
wanghaoran1988: hmm, must forgot that during code clean up.

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -0,0 +1,768 @@
...
+	}
+
+	return true, nil
+}
+
+// This will remove the extra worker nodes we added before kick off upgrade
+func RemoveExtraScaledNodes(c client.Client, upgradeConfig *upgradev1alpha1.UpgradeConfig, reqLogger logr.Logger) (bool, error) {
+	upgradeMachinesets := &machineapi.MachineSetList{}
+
+	err := c.List(context.TODO(), upgradeMachinesets, []client.ListOption{
```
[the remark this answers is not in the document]
wanghaoran1988: Yes, and the nodes are immediately gone.

pkg/controller/upgradeconfig/cluster_upgrader.go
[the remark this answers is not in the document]
wanghaoran1988: that's a good question, do you think we should remove that step for now? and we can revisit if we have a better solution in case of we do need post updates in future?

pkg/controller/upgradeconfig/cluster_upgrader.go
wanghaoran1988: I am not sure, if we need verify that, maybe we can do that in https://github.com/openshift/managed-cluster-validating-webhooks/

pkg/controller/upgradeconfig/cluster_upgrader.go
wanghaoran1988: We are doing step by step upgrade, when we finished step 1, and running step2, and when the 2nd time reconcile starts, we cannot go back check whether step1 finished or not based on the current cluster status, that's why we so much rely on the `conditions` here.

pkg/controller/upgradeconfig/cluster_upgrader.go
wanghaoran1988: @sedroche I don't have a good answer here, could you create a separate issue discuss this? in my mind, if the process already started, it's `upgrading` in our own upgrade process.

pkg/controller/upgradeconfig/cluster_upgrader.go
sedroche: In my experience this always ends up being a problem eventually. If an upgrade goes wrong and a sre needs to restart the upgrade from a certain point this can stop the operator from recovering correctly (bar manually updating the status)

pkg/controller/upgradeconfig/cluster_upgrader.go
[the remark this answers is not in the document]
sedroche: This is definitely not validation though. If we are updating the channel lets update it in the `CommenceUpgrade` step.

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -0,0 +1,768 @@
...
+	Status string    `json:"status"`
+	Data   AlertData `json:"data"`
+}
+
+type AlertData struct {
+	Result []interface{} `json:"result"`
+}
+
+// validateUpgradeConfig will validate the UpgradeConfig, the desired version should be grater than or equal to the current version
+func ValidateUpgradeConfig(c client.Client, upgradeConfig *upgradev1alpha1.UpgradeConfig, reqLogger logr.Logger) (result bool, err error) {
```
sedroche, line 654: @mrbarge @dofinn @wanghaoran1988  Given all our other validation is in webhooks should this be moved there as well?

pkg/controller/upgradeconfig/cluster_upgrader.go
sedroche: Ok gotcha. This is the difference between our upgradeconfig reconciling and the actual upgrade. Its not a big deal so just ignore if you want but is `Pending` a better fit here?

pkg/controller/upgradeconfig/cluster_upgrader.go
mrbarge: @jroche:

- Re: checking daemonsets/replicasets - originally I was thinking that it's not necessary, however we still do have some things sitting in the openshift* namespace that aren't managed by operators (webhook server). If everything was a cluster operator I would agree that it's superfluous.

- Re: validate UpgradeConfig via webhooks - completely agree, that is a perfect scenario for it.

- Re: terminology of 'control plane', if we go by this page [0] then I would say that maaaybe they are part of it?

[0] https://docs.openshift.com/container-platform/4.4/architecture/control-plane.html

wanghaoran1988: https://issues.redhat.com/browse/OSD-3915 created

pkg/controller/upgradeconfig/cluster_upgrader.go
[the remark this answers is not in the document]
wanghaoran1988: Without apply this, we cannot get the availableUpgrades from clusterversion, I will add a TODO to this func for future discussion(move validation to https://github.com/openshift/managed-cluster-validating-webhooks)

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -0,0 +1,768 @@
...
+
+		return false, fmt.Errorf("failed to get current version")
+	}
+	// If the version match, it means it's already upgraded or at least control plane is upgraded.
+	if current == upgradeConfig.Spec.Desired.Version {
+		log.Info("the expected version match current version")
+		return false, fmt.Errorf("cluster is already on version %s", current)
+	}
+
+	// Compare the versions, if the current version is greater than desired, failed the validation
```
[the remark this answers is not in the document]
wanghaoran1988: Will add a comment for sure

pkg/controller/upgradeconfig/cluster_upgrader.go
wanghaoran1988: Let's use this https://issues.redhat.com/browse/OSD-3915

pkg/controller/upgradeconfig/cluster_upgrader.go
wanghaoran1988: It's a lot more work than expected without using this, let's address this problem in a follow up jira, I will create one and linked here

wanghaoran1988: https://issues.redhat.com/browse/OSD-3916

pkg/controller/upgradeconfig/cluster_upgrader.go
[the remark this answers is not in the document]
bmeng: The master upgrade should include two parts, the control plane services(operators, include the api, controller-manager, etcd, etc..) and the master node(coreos, kubelet, etc..) 
The services are done by CVO, and the node (OS level) itself done by MCO.

pkg/controller/upgradeconfig/cluster_upgrader.go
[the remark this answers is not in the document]
sedroche: I still think this is not ideal. We are setting a - possibly incorrect - channel on this ClusterVersion and then leaving the ClusterVersion in that state.
I think we need to consider investigating what CVO does and replicating it.
https://github.com/openshift/cluster-version-operator/blob/b9a3ffafdc32e534c68112a13b89cf1459e1bbd9/pkg/cvo/availableupdates.go#L169
https://github.com/openshift/cluster-version-operator/blob/b9a3ffafdc32e534c68112a13b89cf1459e1bbd9/pkg/cincinnati/cincinnati.go#L23

wdyt? (follow up ticket again is ok)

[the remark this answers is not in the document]
wanghaoran1988: hmm, using cincinnati client to get the available updates look good to me, I will remove the code here, and create a followup ticket.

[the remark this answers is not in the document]
wanghaoran1988: @sedroche I updated the code to use cincinnati client, PTAL

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -0,0 +1,737 @@
...
+	configPool := &machineconfigapi.MachineConfigPool{}
+	err := c.Get(context.TODO(), types.NamespacedName{Name: nodeType}, configPool)
+	if err != nil {
+		return false, nil
+	}
+	//TODO send timeout alert if wait timeout
+	if configPool.Status.MachineCount != configPool.Status.UpdatedMachineCount {
+		errMsg := fmt.Sprintf("not all %s are upgraded, upgraded: %v, totall: %v", nodeType, configPool.Status.UpdatedMachineCount, configPool.Status.MachineCount)
+		reqLogger.Info(errMsg)
+		return false, fmt.Errorf(errMsg)
```
mrbarge, line 406: This results in some pretty spammy error logging [0] during the node upgrades - but it's not really an "error" that some nodes still aren't yet upgraded. Thoughts on maybe just not returning an error at all here? The existing info logging is enough.

```suggestion
		return false, nil
```

[0] {"level":"error","ts":1591072217.1054864,"logger":"controller_upgradeconfig","msg":"error when AllWorkerNodesUpgraded","Request.Namespace":"","Request.Name":"example-upgrade-config","error":"no
t all worker are upgraded, upgraded: 6, totall: 8","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/home/mbargenq/go/pkg/mod/github.com/go-logr/zapr@v0.1.1/zapr.go:128\ngithub.com/o
penshift/managed-upgrade-operator/pkg/controller/upgradeconfig.ClusterUpgrader.UpgradeCluster\n\tmanaged-upgrade-operator/pkg/controller/upgradeconfig/cluster_upgrader.go:464\ngithub.com/openshift/managed-upgrade-operator/pkg/controller/upgradeconfig.(*ReconcileUpgradeConfig).Reconcile\n\tmanaged-upgrade-operator/pkg/controller/upgradeconfig/upgradeconfig_controller.go:143\nsigs.k8s.
io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/home/mbargenq/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.5.2/pkg/internal/controller/controller.go:256\nsigs.k
8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/home/mbargenq/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.5.2/pkg/internal/controller/controller.go:232\n
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker\n\t/home/mbargenq/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.5.2/pkg/internal/controller/controller.go:211\nk8s.io/
apimachinery/pkg/util/wait.JitterUntil.func1\n\t/home/mbargenq/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20191004115801-a2eda9f80ab8/pkg/util/wait/wait.go:152\nk8s.io/apimachinery/pkg/util/wait.Jit
terUntil\n\t/home/mbargenq/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20191004115801-a2eda9f80ab8/pkg/util/wait/wait.go:153\nk8s.io/apimachinery/pkg/util/wait.Until\n\t/home/mbargenq/go/pkg/mod/k8s.
io/apimachinery@v0.0.0-20191004115801-a2eda9f80ab8/pkg/util/wait/wait.go:88"}

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -0,0 +1,737 @@
...
+	for _, ms := range upgradeMachinesets.Items {
+		//We assume the create time is the start time for scale up extra compute nodes
+		startTime := ms.CreationTimestamp
+		if ms.Status.Replicas != ms.Status.ReadyReplicas {
+
+			if time.Now().After(startTime.Time.Add(TIMEOUT_SCALE_EXTRAL_NODES)) {
+				//TODO send out timeout alerts
+				log.Info(fmt.Sprintf("machineset provisioning timout"))
+			}
+			log.Info(fmt.Sprintf("not all mahines are ready for machineset:%s", ms.Name))
```
mrbarge, line 185: Just going through and fixing some small typos in logging.

```suggestion
			log.Info(fmt.Sprintf("not all machines are ready for machineset:%s", ms.Name))
```

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -0,0 +1,737 @@
...
+// Check whether nodes are upgraded or not
+func NodesUpgraded(c client.Client, nodeType string, reqLogger logr.Logger) (bool, error) {
+	configPool := &machineconfigapi.MachineConfigPool{}
+	err := c.Get(context.TODO(), types.NamespacedName{Name: nodeType}, configPool)
+	if err != nil {
+		return false, nil
+	}
+	//TODO send timeout alert if wait timeout
+	if configPool.Status.MachineCount != configPool.Status.UpdatedMachineCount {
+		errMsg := fmt.Sprintf("not all %s are upgraded, upgraded: %v, totall: %v", nodeType, configPool.Status.UpdatedMachineCount, configPool.Status.MachineCount)
```
mrbarge, line 404: Fixing small typos.

```suggestion
		errMsg := fmt.Sprintf("not all %s are upgraded, upgraded: %v, total: %v", nodeType, configPool.Status.UpdatedMachineCount, configPool.Status.MachineCount)
```

pkg/controller/upgradeconfig/upgradeconfig_controller.go
```diff
@@ -89,7 +93,93 @@ func (r *ReconcileUpgradeConfig) Reconcile(request reconcile.Request) (reconcile
...
+
+		} else {
+			r.updateStatusPending(reqLogger, instance)
+			return reconcile.Result{}, nil
+		}
+	case upgradev1alpha1.UpgradePhaseUpgrading:
+		reqLogger.Info("it's upgrading now")
+		upgrader.UpgradeCluster(r.client, instance, reqLogger)
+	case upgradev1alpha1.UpgradePhaseUpgraded:
+		reqLogger.Info("cluster is already already upgraded")
```
mrbarge, line 145: Fixing small typos.

```suggestion
		reqLogger.Info("cluster is already upgraded")
```

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -0,0 +1,737 @@
...
+
+			}
+
+		}
+		if !nodeReady {
+			allNodeReady = false
+			if time.Now().After(startTime.Time.Add(TIMEOUT_SCALE_EXTRAL_NODES)) {
+				log.Info("node is not ready within 30mins")
+				//TODO send out timeout alerts
+				return false, fmt.Errorf("timeout waiting node:%s become ready", nodeName)
```
mrbarge, line 216: Fixing small grammatical issues
```suggestion
				return false, fmt.Errorf("timeout waiting for node:%s to become ready", nodeName)
```

pkg/controller/upgradeconfig/cluster_upgrader.go
sedroche: +1 we want our logging spot on for this

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -0,0 +1,768 @@
...
+	url := "https://" + route.Spec.Host + "/api/v1/query"
+
+	req, err := http.NewRequest("GET", url, nil)
+	q := req.URL.Query()
+	alertQuery := "ALERTS{alertstate=\"firing\",severity=\"critical\",namespace=~\"^openshift.*|^kube.*|^default$\",namespace!=\"openshift-customer-monitoring\",alertname!=\"ClusterUpgradingSRE\",alertname!=\"DNSErrors05MinSRE\",alertname!=\"MetricsClientSendFailingSRE\"}"
+	q.Add("query", alertQuery)
+
+	req.URL.RawQuery = q.Encode()
+	req.Header.Add("Authorization", "Bearer "+string(token))
+	resp, err := hclient.Do(req)
```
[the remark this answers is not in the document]
sedroche: @wanghaoran1988 The prometheus operator has a client -> https://github.com/coreos/prometheus-operator/blob/master/pkg/prometheus/operator.go#L36 might be worth a look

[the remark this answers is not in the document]
wanghaoran1988: @sedroche  https://issues.redhat.com/browse/OSD-3931 to track this

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -0,0 +1,737 @@
...
+	if err != nil {
+		log.Info(fmt.Sprintf("failed to get clusterversion"))
+		log.Error(err, "failed to get clusterversion")
+		return false, err
+	}
+
+	//TODO get available version from ocm api like : ocm get "https://api.openshift.com/api/clusters_mgmt/v1/versions" --parameter search="enabled='t'"
+
+	//Get current version, then compare
+	current := getCurrentlVersion(clusterVersion)
```
mrbarge, line 640: Minor typo suggestion, can we rename this to "getCurrentVersion"

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -0,0 +1,737 @@
...
+		cvoUpdates = append(cvoUpdates, configv1.Update{
+			Version: update.Version.String(),
+			Image:   update.Image,
+		})
+	}
+	// Check whether the desired version exists in availableUpdates
+
+	found := false
+	for _, v := range cvoUpdates {
+		if v.Version == upgradeConfig.Spec.Desired.Version && v.Force == false {
```
mrbarge, line 692: I'm curious why the "vForce == false" here? Not objecting to it, just want to understand why.

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -0,0 +1,737 @@
...
+
+	token := secret.Data[corev1.ServiceAccountTokenKey]
+
+	route := &routev1.Route{}
+	err = c.Get(context.TODO(), types.NamespacedName{Namespace: "openshift-monitoring", Name: "prometheus-k8s"}, route)
+	if err != nil {
+		return false, err
+	}
+	tr := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
```
sedroche, line 555: I see the docs for this say:
```
    // InsecureSkipVerify controls whether a client verifies the
    // server's certificate chain and host name.
    // If InsecureSkipVerify is true, TLS accepts any certificate
    // presented by the server and any host name in that certificate.
    // In this mode, TLS is susceptible to man-in-the-middle attacks.
    // This should be used only for testing.
    InsecureSkipVerify bool
```

Is this left over from testing?

pkg/controller/upgradeconfig/cluster_upgrader.go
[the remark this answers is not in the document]
sedroche: awesome

pkg/controller/upgradeconfig/cluster_upgrader.go
wanghaoran1988: > 	// force allows an administrator to update to an image that has failed
> 	// verification, does not appear in the availableUpdates list, or otherwise
> 	// would be blocked by normal protections on update. This option should only
> 	// be used when the authenticity of the provided image has been verified out
> 	// of band because the provided image will run with full administrative access
> 	// to the cluster. Do not use this flag with images that comes from unknown
> 	// or potentially malicious sources.
> 	//
> 	// This flag does not override other forms of consistency checking that are
> 	// required before a new update is deployed.

This is the description for the `False` congifuration.

pkg/controller/upgradeconfig/cluster_upgrader.go
wanghaoran1988: Cluster installed in `int` env, are using self signed ca, see comments here: https://github.com/openshift/managed-upgrade-operator/pull/9#discussion_r432184617

pkg/controller/upgradeconfig/cluster_upgrader.go
mrbarge: Thanks, that now makes perfect sense. Cheers!

pkg/controller/upgradeconfig/cluster_upgrader.go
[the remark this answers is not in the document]
sedroche: If we create the MachineSet with a deterministic name can we just do a straight delete?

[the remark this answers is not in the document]
wanghaoran1988: We are not saving the names somewhere, we may have more machineset created if it's multi-AZ.

pkg/controller/upgradeconfig/cluster_upgrader.go
wanghaoran1988: https://issues.redhat.com/browse/OSD-3945 created as a followup

PR #10, opened 2020-06-02
AlertManager based maintenace API

sedroche:
https://issues.redhat.com/browse/OSD-3658

This is what im thinking the maintenance client will look like. (I'll fit it into [the upgrade controller](https://github.com/openshift/managed-upgrade-operator/pull/9) once that is ready)

Should be just:

```.go
m, err := maintenance.NewMaintenace()
  if err != nil {
  }
   err = m.Start()
     /
     /
  err = m.End()
```

### DIFF
--- a/pkg/controller/upgradeconfig/cluster_upgrader.go
+++ b/pkg/controller/upgradeconfig/cluster_upgrader.go
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
+	"github.com/openshift/managed-upgrade-operator/pkg/maintenance"
 	"io/ioutil"
 	"net/http"
 	"net/url"
@@ -68,7 +69,7 @@ func Ordering() []upgradev1alpha1.UpgradeConditionType {
 	return UpgradeStepOrdering
 }
 
-type UpgradeStep func(client client.Client, upgradeConfig *upgradev1alpha1.UpgradeConfig, reqLogger logr.Logger) (bool, error)
+type UpgradeStep func(c client.Client, m maintenance.Maintenance, upgradeConfig *upgradev1alpha1.UpgradeConfig, reqLogger logr.Logger) (bool, error)
 
 func NewUpgrader() ClusterUpgrader {
 	once.Do(func() {
@@ -95,12 +96,12 @@ func NewUpgrader() ClusterUpgrader {
 }
 
 // ClusterHealthCheck performs cluster healthy check
-func PreClusterHealthCheck(c client.Client, upgradeConfig *upgradev1alpha1.UpgradeConfig, reqLogger logr.Logger) (bool, error) {
+func PreClusterHealthCheck(c client.Client, m maintenance.Maintenance, upgradeConfig *upgradev1alpha1.UpgradeConfig, reqLogger logr.Logger) (bool, error) {
 	return performClusterHealthCheck(c)
 }
 
 // This will create a new machineset with 1 extra replicas for workers in every region
-func EnsureExtraUpgradeWorkers(c client.Client, upgradeConfig *upgradev1alpha1.UpgradeConfig, reqLogger logr.Logger) (bool, error) {
+func EnsureExtraUpgradeWorkers(c client.Client, m maintenance.Maintenance, upgradeConfig *upgradev1alpha1.UpgradeConfig, reqLogger logr.Logger) (bool, error) {
 	upgradeMachinesets := &machineapi.MachineSetList{}
 
 	err := c.List(context.TODO(), upgradeMachinesets, []client.ListOption{
@@ -228,7 +229,7 @@ func EnsureExtraUpgradeWorkers(c client.Client, upgradeConfig *upgradev1alpha1.U
 }
 
 // CommenceUpgrade will update the clusterversion object to apply the desired version to trigger real OCP upgrade
-func CommenceUpgrade(c client.Client, upgradeConfig *upgradev1alpha1.UpgradeConfig, reqLogger logr.Logger) (bool, error) {
+func CommenceUpgrade(c client.Client, m maintenance.Maintenance, upgradeConfig *upgradev1alpha1.UpgradeConfig, reqLogger logr.Logger) (bool, error) {
 	clusterVersion := &configv1.ClusterVersion{}
 	err := c.Get(context.TODO(), types.NamespacedName{Name: "version"}, clusterVersion)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/maintenance/maintenance.go
```diff
@@ -0,0 +1,27 @@
...
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type Maintenance interface {
+	Start() error
+	End() error
+}
+
+func NewMaintenace() (Maintenance, error) {
+	// new client to avoid caching issues
```
sedroche, line 14: Actually @wanghaoran1988 why do I have to do this?
I need a non-cached client to get this route. any ideas?

wanghaoran1988: Could you point me why we need the non-cached client?

bmeng:
should we add comments to each function?

pkg/maintenance/maintenance.go
sedroche: The cached client wasn't retrieving the alertmanager route.
I switched to the non-cached and it was retrieved. I will see if the cached works once I merged into your branch but was wondering if you had seen this

sedroche:
> should we add comments to each function?

Will do @bmeng. thanks

pkg/maintenance/maintenance.go
wanghaoran1988: @sedroche How you deploy the operator, if it's by a deployment objet, the following must be set:
```
 - name: WATCH_NAMESPACE
   value: ""
```
or you test locally with:
`operator-sdk run --local --verbose --watch-namespace ""`

pkg/maintenance/alertmanager.go
```diff
@@ -0,0 +1,173 @@
...
+
+func (amm *alertManagerMaintenance) Start() error {
+	createdBy := "managed-upgrade-operator"
+	comment := "Test comment"
+	startsAt := strfmt.DateTime(time.Now().UTC())
+	endsAt := strfmt.DateTime(time.Now().Add(time.Minute * 5))
+
+	name := "alertname"
+	isRegex := true
+	value := "[A-Z].*"
```
wanghaoran1988, line 95: We need make sure this only silence the alerts not upgrade related

mrbarge: I agree, was going to make the same comment. Can we try and make it exclude the existing "ClusterUpgradingSRE" alert from the silence? And maybe add more later if we need to.

sedroche: Yeah i'll refine these when I rebase on haorans work :+1:

pkg/maintenance/maintenance.go
sedroche: Just locally from goland. Have that set in my environemnt too. weird.

pkg/maintenance/alertmanager.go
```diff
@@ -0,0 +1,173 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (amm *alertManagerMaintenance) End() error {
+	dParams := &amSilence.DeleteSilenceParams{
+		SilenceID:  "",
```
mrbarge, line 127: Just curious (I don't know the answer) - does having no SilenceIDs specified result in deleting all silences, or will End() need to query for the existing silence IDs first in order to populate this?

sedroche: The api needs an actual id for each delete.
Once I start integrating it ill see do we need to store an id or maybe just do a get and check `CreatedBy` = "managed-upgrade-operator" or whatever.

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -250,40 +251,61 @@ func CommenceUpgrade(c client.Client, upgradeConfig *upgradev1alpha1.UpgradeConf
 	return true, nil
 }
 
-// TODO This create the maint window for control plane
-func CreateControlPlaneMaintWindow(c client.Client, upgradeConfig *upgradev1alpha1.UpgradeConfig, reqLogger logr.Logger) (bool, error) {
+// Create the maintenance window for control plane
+func CreateControlPlaneMaintWindow(c client.Client, m maintenance.Maintenance, upgradeConfig *upgradev1alpha1.UpgradeConfig, reqLogger logr.Logger) (bool, error) {
+	err := m.Start(time.Now().Add(time.Minute * 90))
```
mrbarge, line 256: I'll raise it now because it's in my mind, but I wonder if there's an opportunity for us to have a better place for storing the static 'control plane' and 'worker' time periods - hidden in code makes it easy to miss - arguably the same reason why the maintenance window logic for the previous upgrade script was not easily discernable.

Not sure about the best way. A configmap for the operator maybe? Or an amendment to the UpgradeConfig CRD. But something I personally think would be a good opportunity to address before we hit Milestone 1. What do you think?

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -250,40 +251,61 @@ func CommenceUpgrade(c client.Client, upgradeConfig *upgradev1alpha1.UpgradeConf
...
+// Create the maintenance window for workers
+func CreateWorkerMaintWindow(c client.Client, m maintenance.Maintenance, upgradeConfig *upgradev1alpha1.UpgradeConfig, reqLogger logr.Logger) (bool, error) {
+	configPool := &machineconfigapi.MachineConfigPool{}
+	err := c.Get(context.TODO(), types.NamespacedName{Name: "worker"}, configPool)
+	if err != nil {
+		return false, nil
+	}
+
+	pendingWorkerCount := configPool.Status.MachineCount - configPool.Status.UpdatedMachineCount
+	var maintenanceTimePerNode int32 = 8
```
mrbarge, line 283: See my earlier comment on line 256. But regardless, for this specific line, would prefer we set it to 
```suggestion
	var maintenanceTimePerNode int32 = 8 * time.Minute
```
here, rather than multiply it by the time.Minute down in the `m.Start()` section - that way it is blatantly obvious when we see it defined what unit of time it represents.

pkg/maintenance/alertManagerSilenceClient.go
```diff
@@ -0,0 +1,73 @@
...
+	silenceClient := amSilence.New(ams.transport, strfmt.Default)
+	results, err := silenceClient.GetSilences(gParams)
+	if err != nil {
+		return nil, err
+	}
+
+	return results, nil
+}
+
+// list silence in Alertmanager instance defined in transport
```
bmeng, line 58: typo

pkg/maintenance/alertManagerSilenceClient.go
```diff
@@ -0,0 +1,73 @@
...
+	"net/http"
+)
+
+type alertManagerSilenceClient struct {
+	transport *httptransport.Runtime
+}
+
+// Creates a silence in Alertmanager instance defined in transport
+func (ams *alertManagerSilenceClient) create(matchers amv2Models.Matchers, startsAt strfmt.DateTime, endsAt strfmt.DateTime, creator string, comment string) error {
+	psParams := &amSilence.PostSilencesParams{
```
bmeng, line 18: should we use `pParams` here as the same format below?

sedroche:
Thanks @bmeng see -> cd957e43a960e33b344098d4a44a38f0eb62f9d9

sedroche:
Thanks @mrbarge see -> e6fe4c34bfd21f0f25db66737e5faf1f5d94aeea

pkg/controller/upgradeconfig/cluster_upgrader.go
sedroche: @mrbarge  Definite +1 to extracting config.
I would go with a configmap but was wondering about how its created/managed. Would this also be an object in the `SelectorSyncSet`?

mrbarge: @jroche yes that's what I'd be thinking. Like, maybe a ConfigMap which is managed as part of the `managed-cluster-config` setup for the operator, which would then get incorporated into a `SelectorSyncSet` when the next release of `managed-cluster-config` is built.

PR #11, opened 2020-06-09
Adding build artifacts for CI

ravitri:
Adding build configuration to onboard repo to Openshift CI as per https://issues.redhat.com/browse/OSD-3781

### DIFF
--- a/project.mk
+++ b/project.mk
@@ -3,7 +3,7 @@ OPERATOR_NAME?=managed-upgrade-operator
 OPERATOR_NAMESPACE?=managed-upgrade-operator
 
 IMAGE_REGISTRY?=quay.io
-IMAGE_REPOSITORY?=openshift-sre
+IMAGE_REPOSITORY?=$(USER)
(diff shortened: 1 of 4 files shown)


project.mk
```diff
@@ -0,0 +1,10 @@
+# Project specific values
+OPERATOR_NAME?=managed-upgrade-operator
+OPERATOR_NAMESPACE?=managed-upgrade-operator
+
+IMAGE_REGISTRY?=quay.io
+IMAGE_REPOSITORY?=openshift-sre
```
mrbarge, line 6: I see other SRE operators (gcp, cloud-ingress, dms, aws) tend to do:
```suggestion
IMAGE_REPOSITORY?=$(USER)
```
here.. which I guess assists when doing personal development. Could we do a similar thing here?

ravitri: Ok, I understand your point. Yeah we could do that as well. The eventual operator deployment image will be from a standard location as such so this could help with dev stuff..

ravitri: Have added the change

PR #17, opened 2020-06-16
updated readme w/dev info

dofinn:


docs/development.md
```diff
@@ -0,0 +1,35 @@
+# Development
+
+This document should entail all you need to develop this operator locally. 
+
+## Versions
+
+### golang
+
+`go version go1.13`
```
sedroche, line 9: \>=1.13

docs/development.md
```diff
@@ -0,0 +1,35 @@
...
+
+### operator-sdk
+
+`operator-sdk version: "v0.17.0", commit: "2fd7019f856cdb6f6618e2c3c80d15c3c79d1b6c", kubernetes version: "v1.17.2", go version: "go1.14.2 linux/amd64"`
```
sedroche, line 13: Feels very specific. Just say v0.17?

docs/development.md
```diff
@@ -0,0 +1,35 @@
...
+
+## How to run
+
+1. Either fork your own or checkout the repo from https://github.com/openshift/managed-upgrade-operator into your working go path. For example:
```
sedroche, line 17: This is no longer true using go modules. repo doesn't need to be in gopath

docs/development.md
ravitri, line 13: Golang version above mentions of 1.13 and the client version in operator-sdk version shows go1.14.2. We could mark it same to avoid confusion?

docs/development.md
```diff
@@ -0,0 +1,35 @@
...
+```
+cd ~/go/src/github.com/openshift
+git clone https://github.com/openshift/managed-upgrade-operator.git
+cd managed-upgrade-operator
+```
+
+2. From within the directory, run:
+
+```
+operator-sdk run --local --watch-namespace ""
```
sedroche, line 28: Maybe a `Make run` command? (up to you)

docs/development.md
dofinn: good catch

docs/development.md
dofinn: yep

PR #18, opened 2020-06-17
Refinement of upgrade maintenance

sedroche:
There was a discussion about this within the monitoring team (https://issues.redhat.com/browse/MON-1093)

tl;dr is that critical alerts "shouldn't" occur during upgrade but that there is a [shortlist of criticals](https://bugzilla.redhat.com/buglist.cgi?quicksearch=%5Balert%5D) than do/may occur during upgrade. (There is work to refine those to not fire/fire less but a target of 4.7 has been set)

- Silencing all warning|info alerts as these can be noise during an upgrade.
- From the above critical list only  `etcdMembersDown` is certain to happen so I have silenced that for control plane maintenance and we can refine this list as we iterate further and see alerts.

### DIFF
--- a/pkg/controller/upgradeconfig/cluster_upgrader.go
+++ b/pkg/controller/upgradeconfig/cluster_upgrader.go
@@ -255,7 +255,7 @@ func CommenceUpgrade(c client.Client, m maintenance.Maintenance, upgradeConfig *
 // Create the maintenance window for control plane
 func CreateControlPlaneMaintWindow(c client.Client, m maintenance.Maintenance, upgradeConfig *upgradev1alpha1.UpgradeConfig, reqLogger logr.Logger) (bool, error) {
 	endTime := time.Now().Add(90 * time.Minute)
-	err := m.Start(endTime)
+	err := m.StartControlPlane(endTime)
 	if err != nil {
 		return false, err
 	}
@@ -285,7 +285,7 @@ func CreateWorkerMaintWindow(c client.Client, m maintenance.Maintenance, upgrade
 	maintenanceDurationPerNode := 8 * time.Minute
 	workerMaintenanceExpectedDuration := time.Duration(pendingWorkerCount) * maintenanceDurationPerNode
 	endTime := time.Now().Add(workerMaintenanceExpectedDuration)
-	err = m.Start(endTime)
+	err = m.StartWorker(endTime)
 	if err != nil {
 		return false, err
 	}
--- a/pkg/maintenance/alertmanagerMaintenance.go
+++ b/pkg/maintenance/alertmanagerMaintenance.go
@@ -21,6 +21,11 @@ var (
 	alertManagerRouteName          = "alertmanager-main"
 	alertManagerServiceAccountName = "prometheus-k8s"
 	alertManagerBasePath           = "/api/v2/"
+
+	// Generally upgrades should not fire critical alerts but there are some critical alerts that will fire.
+	// e.g. 'etcdMembersDown' happens as the masters drain/reboot and a master is offline but this is expected and will resolve.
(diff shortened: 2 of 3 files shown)


pkg/maintenance/alertmanagerMaintenance.go
```diff
@@ -21,6 +21,11 @@ var (
 	alertManagerRouteName          = "alertmanager-main"
 	alertManagerServiceAccountName = "prometheus-k8s"
 	alertManagerBasePath           = "/api/v2/"
+
+	// Generally upgrades should not fire critical alerts but there are some critical alerts that will fire.
+	// e.g. 'etcdMembersDown' happens as the masters drain/reboot and a master is offline but this is expected and will resolve.
+	// This is a regex of critical alerts that can be ignored while upgrading of controlplane occurs
+	controlPlaneIgnoredCriticalAlerts = "(etcdMembersDown)"
```
sedroche, line 28: Just calling out that this is silencing a etcd member down for potentially 90 mins but I don't know any way to refine it further other than what is called out in https://bugzilla.redhat.com/show_bug.cgi?id=1824986 which is probably to wait to alert after a certain amount of time down.

pkg/maintenance/alertmanagerMaintenance.go
```diff
@@ -123,3 +141,19 @@ func (amm *alertManagerMaintenance) End() error {
...
+	return &amv2Models.Matcher{
+		Name:    &alertMatchKey,
+		IsRegex: &isRegex,
+		Value:   &alertValue,
+	}
+}
+
+func createDefaultMatchers() []*amv2Models.Matcher {
+	// Upgrades can impact some availability which may trigger info/warning alerts. ignore those.
+	nonCriticalAlertMatcher := createMatcher("severity", "(warning|info|none)", true)
```
sedroche, line 155: @bmeng fyi since we are only silencing on `warning|info|none` then your alerts can just be critical severity to alert SRE

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -255,7 +255,7 @@ func CommenceUpgrade(c client.Client, m maintenance.Maintenance, upgradeConfig *
 // Create the maintenance window for control plane
 func CreateControlPlaneMaintWindow(c client.Client, m maintenance.Maintenance, upgradeConfig *upgradev1alpha1.UpgradeConfig, reqLogger logr.Logger) (bool, error) {
 	endTime := time.Now().Add(90 * time.Minute)
```
mrbarge, line 257: Not a commit specific to your PR changes, but, they made me think of it.
In the maintenance client you explicitly convert to UTC with "time.Now().UTC()" when generating the 'now' timestamp.
For consistency should we do the same when generating the endTime?
(It's a minor nit and yeah all our servers are on UTC)

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -285,7 +285,7 @@ func CreateWorkerMaintWindow(c client.Client, m maintenance.Maintenance, upgrade
 	maintenanceDurationPerNode := 8 * time.Minute
 	workerMaintenanceExpectedDuration := time.Duration(pendingWorkerCount) * maintenanceDurationPerNode
 	endTime := time.Now().Add(workerMaintenanceExpectedDuration)
```
mrbarge, line 287: see also: my earlier comment

pkg/controller/upgradeconfig/cluster_upgrader.go
sedroche: Nice catch @mrbarge 
Check 2c192d67cacd8bdf7202c439afbde1f97d1439a0

PR #22, opened 2020-06-19
add the metrics to the operator to track the upgrade status

bmeng:
With the metrics added, we can create the prometheus rules to trigger the alerts.

Prometheus rules:
1. validation failed
2. cluster check failed
3. upgrade started for 90 mins and controlplane upgrade not finished
4. upgrade started for 3 hours and node upgrade not finished 
5. Post verification failed

With rules like:
```
apiVersion: monitoring.coreos.com/v1
kind: PrometheusRule
metadata:
  annotations:
  labels:
    prometheus: managed-upgrade-operator
    role: alert-rules
  name: sre-managed-upgrade-alerts
  namespace: openshift-monitoring
spec:
  groups:
  - name: sre-managed-upgrade-alerts
    rules:
    - alert: UpgradeConfigValidationFailed
      expr: upgradeoperator_upgradeconfig_validation_failed > 0
      labels:
        severity: critical
    - alert: UpgradeClusterHealthCheckFailed
      expr: upgradeoperator_cluster_check_failed > 0
      labels:
        severity: critical
    - alert: UpgradeControlPlaneTimeout
      expr:  //PromQL needs to be worked out
      for: 1m
      labels:
        severity: critical
    - alert: UpgradeNodeTimeout
      expr:  //PromQL needs to be worked out
      for: 1m
      labels:
        severity: critical
    - alert: UpgradePostValidationFailed
      expr: upgradeoperator_post_verification_failed > 0
      labels: 
        severity: critical
```

### DIFF
--- a/pkg/controller/upgradeconfig/cluster_upgrader.go
+++ b/pkg/controller/upgradeconfig/cluster_upgrader.go
@@ -247,7 +247,7 @@ func CommenceUpgrade(c client.Client, m maintenance.Maintenance, upgradeConfig *
 	clusterVersion.Spec.Channel = upgradeConfig.Spec.Desired.Channel
 
 	//Record the timestamp when we start the upgrade
-	metrics.UpdateMetricUpgradeStartTime(time.Now(), upgradeConfig.Name)
+	metrics.UpdateMetricUpgradeStartTime(time.Now().UTC(), upgradeConfig.Name)
 	err = c.Update(context.TODO(), clusterVersion)
 	if err != nil {
 		return false, err
@@ -441,7 +441,7 @@ func NodesUpgraded(c client.Client, nodeType string, upgradeConfig *upgradev1alp
 	}
 
 	// send node upgrade complete metrics
-	metrics.UpdateMetricNodeUpgradeEndTime(time.Now(), upgradeConfig.Name)
+	metrics.UpdateMetricNodeUpgradeEndTime(time.Now().UTC(), upgradeConfig.Name)
 	return true, nil
 }
 
@@ -456,7 +456,7 @@ func ControlPlaneUpgraded(c client.Client, m maintenance.Maintenance, upgradeCon
 	for _, c := range clusterVersion.Status.History {
 		if c.State == configv1.CompletedUpdate && c.Version == upgradeConfig.Spec.Desired.Version {
 			// send controlplane upgrade complete timestamp
-			metrics.UpdateMetricControlPlaneEndTime(time.Now(), upgradeConfig.Name)
+			metrics.UpdateMetricControlPlaneEndTime(time.Now().UTC(), upgradeConfig.Name)
 			return true, nil
 		}
 	}
--- a/pkg/controller/upgradeconfig/cluster_upgrader.go
+++ b/pkg/controller/upgradeconfig/cluster_upgrader.go
@@ -5,6 +5,8 @@ import (
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
+	"github.com/openshift/managed-upgrade-operator/pkg/maintenance"
+	"github.com/openshift/managed-upgrade-operator/pkg/metrics"
 	"io/ioutil"
 	"net/http"
 	"net/url"
@@ -13,8 +15,6 @@ import (
 	"strings"
 	"sync"
 	"time"
-
-	"github.com/openshift/managed-upgrade-operator/pkg/maintenance"
 
 	"github.com/blang/semver"
 	"github.com/go-logr/logr"
@@ -98,7 +98,7 @@ func NewUpgrader() ClusterUpgrader {
 
 // ClusterHealthCheck performs cluster healthy check
 func PreClusterHealthCheck(c client.Client, m maintenance.Maintenance, upgradeConfig *upgradev1alpha1.UpgradeConfig, reqLogger logr.Logger) (bool, error) {
-	return performClusterHealthCheck(c)
+	return performClusterHealthCheck(c, upgradeConfig)
 }
 
 // This will create a new machineset with 1 extra replicas for workers in every region
@@ -245,6 +245,9 @@ func CommenceUpgrade(c client.Client, m maintenance.Maintenance, upgradeConfig *
 	clusterVersion.Spec.Overrides = []configv1.ComponentOverride{}
 	clusterVersion.Spec.DesiredUpdate = &configv1.Update{Version: upgradeConfig.Spec.Desired.Version}
 	clusterVersion.Spec.Channel = upgradeConfig.Spec.Desired.Channel
+
+	//Record the timestamp when we start the upgrade
+	metrics.UpdateMetricUpgradeStartTime(time.Now(), upgradeConfig.Name)
 	err = c.Update(context.TODO(), clusterVersion)
 	if err != nil {
 		return false, err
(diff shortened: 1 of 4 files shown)


mrbarge:
Can you check your go.mod to see if you're missing a dependency?

I see the golangci-lint job is failing in the PR pipeline - which it also does for me when I run it locally on your branch.

However, if I run a `go test` job on your branch, it introduces the following new dependency into go.mod:

`github.com/prometheus/client_golang v1.5.1`

and after which the golangci-lint job succeeds.

So are we maybe missing that dependency in your PR?

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -245,6 +245,9 @@ func CommenceUpgrade(c client.Client, m maintenance.Maintenance, upgradeConfig *
 	clusterVersion.Spec.Overrides = []configv1.ComponentOverride{}
 	clusterVersion.Spec.DesiredUpdate = &configv1.Update{Version: upgradeConfig.Spec.Desired.Version}
 	clusterVersion.Spec.Channel = upgradeConfig.Spec.Desired.Channel
+
+	//Record the timestamp when we start the upgrade
+	metrics.UpdateMetricUpgradeStartTime(time.Now(), upgradeConfig.Name)
```
mrbarge, line 250: In the maintenance client, we explicitly get the time as UTC when dealing with it. Thoughts on maybe doing the same here so that we're consistent across all the code? eg.

```suggestion
	metrics.UpdateMetricUpgradeStartTime(time.Now().UTC(), upgradeConfig.Name)
```

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -418,23 +424,24 @@ func RemoveMaintWindow(c client.Client, m maintenance.Maintenance, upgradeConfig
...
 	}
-	//TODO send timeout alert if wait timeout
 	if configPool.Status.MachineCount != configPool.Status.UpdatedMachineCount {
 		errMsg := fmt.Sprintf("not all %s are upgraded, upgraded: %v, total: %v", nodeType, configPool.Status.UpdatedMachineCount, configPool.Status.MachineCount)
 		reqLogger.Info(errMsg)
 		return false, nil
 	}
 
+	// send node upgrade complete metrics
+	metrics.UpdateMetricNodeUpgradeEndTime(time.Now(), upgradeConfig.Name)
```
mrbarge, line 444: See my comment elsewhere in the review.
```suggestion
	metrics.UpdateMetricNodeUpgradeEndTime(time.Now().UTC(), upgradeConfig.Name)
```

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -448,6 +455,8 @@ func ControlPlaneUpgraded(c client.Client, m maintenance.Maintenance, upgradeCon
 	}
 	for _, c := range clusterVersion.Status.History {
 		if c.State == configv1.CompletedUpdate && c.Version == upgradeConfig.Spec.Desired.Version {
+			// send controlplane upgrade complete timestamp
+			metrics.UpdateMetricControlPlaneEndTime(time.Now(), upgradeConfig.Name)
```
mrbarge, line 459: See my comment elsewhere in the review.
```suggestion
			metrics.UpdateMetricControlPlaneEndTime(time.Now().UTC(), upgradeConfig.Name)
```

sedroche:
:eyes:

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -418,23 +424,24 @@ func RemoveMaintWindow(c client.Client, m maintenance.Maintenance, upgradeConfig
 // This perform cluster health check after upgrade
 func PostClusterHealthCheck(c client.Client, m maintenance.Maintenance, upgradeConfig *upgradev1alpha1.UpgradeConfig, reqLogger logr.Logger) (bool, error) {
 
-	return performClusterHealthCheck(c)
+	return performClusterHealthCheck(c, upgradeConfig)
 }
 
 // Check whether nodes are upgraded or not
-func NodesUpgraded(c client.Client, nodeType string, reqLogger logr.Logger) (bool, error) {
+func NodesUpgraded(c client.Client, nodeType string, upgradeConfig *upgradev1alpha1.UpgradeConfig, reqLogger logr.Logger) (bool, error) {
```
dofinn, line 431: Do we need to pass the entire UpgradeConfig object into this func when we only reference a string ? `upgradeConfig.Name`

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -548,7 +557,7 @@ func (cu ClusterUpgrader) UpgradeCluster(c client.Client, m maintenance.Maintena
 // check several things about the cluster and report problems
 // * critical alerts
 // * degraded operators (if there are critical alerts only)
-func performClusterHealthCheck(c client.Client) (bool, error) {
+func performClusterHealthCheck(c client.Client, upgradeConfig *upgradev1alpha1.UpgradeConfig) (bool, error) {
```
dofinn, line 560: Do we need to pass the entire UpgradeConfig object into this func when we only reference a string ? `upgradeConfig.Name`

bmeng: I am not sure how to pass the string to the func else, any suggestion?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,90 @@
...
+func init() {
+	metrics.Registry.MustRegister(metricValidationFailed)
+	metrics.Registry.MustRegister(metricClusterCheckFailed)
+	metrics.Registry.MustRegister(metricUpgradeStartTime)
+	metrics.Registry.MustRegister(metricControlPlaneUpgradeEndTime)
+	metrics.Registry.MustRegister(metricNodeUpgradeEndTime)
+	metrics.Registry.MustRegister(metricPostVerificationFailed)
+}
+
+func UpdateMetricValidationFailed(x int, upgradeconfig string) {
```
sedroche, line 56: If we make these functions
`SetMetricValidationFailed` and `SetMetricValidationPassed` 
`SetMetricPostVerificationFailed` `SetMetricPostVerificationPassed`

etc.

we can avoid passing the int param. Might be easier api. wdyt?

sedroche:
Its not your code but can your promql client do this any easier?
https://github.com/openshift/managed-upgrade-operator/blob/master/pkg/controller/upgradeconfig/cluster_upgrader.go#L552-L625
We have a card for it -> https://issues.redhat.com/browse/OSD-3931

bmeng:
> Its not your code but can your promql client do this any easier?
> https://github.com/openshift/managed-upgrade-operator/blob/master/pkg/controller/upgradeconfig/cluster_upgrader.go#L552-L625
> We have a card for it -> https://issues.redhat.com/browse/OSD-3931

Yes, I think it would be better to use the prometheus client to do that and it is already on my list.

bmeng:
> 1 point on -> "upgrade started for 90 mins and controlplane upgrade not finished"
> The controlplane maintenance will be configurable from a configmap so the promql may be very hard (im not sure how hard not really an expert on alerting)
> Would an alert on whether the controlplane silence has expired work do you think?

Yes, it will pretty hard for PromQL. 
And the initial plan is we set a fixed value for the controlplane and all the node upgrade time, eg, 2 hours for control plane, and 4 hours for all the nodes.
Which are pretty large which the upgrade process should be finished within. If not, we will get the alerts.

I think it is not quite easy to get the alertmanger silence status from prometheus, so maybe cannot use the slience end time.

sedroche:
@bmeng Im not sure if oit is the answer but there is an alert about silences already
https://github.com/openshift/managed-cluster-config/blob/master/deploy/sre-prometheus/100-alertmanager-silence-active.PrometheusRule.yaml

pkg/controller/upgradeconfig/cluster_upgrader.go
```diff
@@ -245,6 +245,9 @@ func CommenceUpgrade(c client.Client, m maintenance.Maintenance, upgradeConfig *
 	clusterVersion.Spec.Overrides = []configv1.ComponentOverride{}
 	clusterVersion.Spec.DesiredUpdate = &configv1.Update{Version: upgradeConfig.Spec.Desired.Version}
 	clusterVersion.Spec.Channel = upgradeConfig.Spec.Desired.Channel
+
+	//Record the timestamp when we start the upgrade
+	metrics.UpdateMetricUpgradeStartTime(time.Now().UTC(), upgradeConfig.Name)
```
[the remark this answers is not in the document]
bmeng: yeah, I think the time which is being used in prometheus should be unix time. Change these back.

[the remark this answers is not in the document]
bmeng: yeah, I think the time which is being used in prometheus should be unix time. Change these back.

PR #25, opened 2020-06-24
refine the metrics package to use inferface

bmeng:


mrbarge:
Looks fine to me! My only minor nit is with the naming of the "ms" variable in cluster_upgrader and the controller. Or, rather, now that a lot of functions have an "m" and a "ms", it presents some minor readability issues going forward. Would prefer the variables be pretty explicit, eg. metricsClient, maintClient. We could alternatively deal with all that as part of the refactoring in OSD-4168.

That's pretty much it from me - will give @sedroche and @dofinn a chance to comment on before adding a lgtm.

pkg/cluster_upgrader/cluster_upgrader.go
```diff
@@ -699,15 +702,15 @@ type AlertData struct {
...
+func ValidateUpgradeConfig(c client.Client, ms metrics.Metrics, m maintenance.Maintenance, upgradeConfig *upgradev1alpha1.UpgradeConfig, reqLogger logr.Logger) (result bool, err error) {
 
 	log.Info("validating upgradeconfig")
 	clusterVersion := &configv1.ClusterVersion{}
 	err = c.Get(context.TODO(), types.NamespacedName{Name: "version"}, clusterVersion)
 	if err != nil {
 		log.Info("failed to get clusterversion")
 		log.Error(err, "failed to get clusterversion")
-		metrics.UpdateMetricValidationFailed(upgradeConfig.Name)
+		ms.UpdateMetricValidationFailed(upgradeConfig.Name)
```
sedroche, line 713: I'd like to avoid calling to metrics within functions if we can. It's harder to maintain (i.e. someone adds new code and doesn't add a call to metrics)
So for `ValidateUpgradeConfig` and `PostUpgradeVerification` can you use the same pattern as for `PostClusterHealthCheck` where we update metrics based on pass/fail of the entire function?

pkg/cluster_upgrader/cluster_upgrader.go
```diff
@@ -320,21 +320,25 @@ func CreateWorkerMaintWindow(c client.Client, m maintenance.Maintenance, upgrade
...
 	return NodesUpgraded(c, "master", upgradeConfig, reqLogger)
 
 }
 
 // This check whether all the worker nodes are ready with new config
-func AllWorkersUpgraded(c client.Client, m maintenance.Maintenance, upgradeConfig *upgradev1alpha1.UpgradeConfig, reqLogger logr.Logger) (bool, error) {
-
-	return NodesUpgraded(c, "worker", upgradeConfig, reqLogger)
+func AllWorkersUpgraded(c client.Client, ms metrics.Metrics, m maintenance.Maintenance, upgradeConfig *upgradev1alpha1.UpgradeConfig, reqLogger logr.Logger) (bool, error) {
+	ok, err := NodesUpgraded(c, "worker", upgradeConfig, reqLogger)
```
sedroche, line 331: Small nitpick. Can you lowercase `NodesUpgraded` -> `nodesUpgraded` and I think it no longer requires the upgrade param

sedroche:
> Would prefer the variables be pretty explicit, eg. metricsClient, maintClient.

I believe it's a go convention to use short var names but I personally prefer the readability of Matts suggestion

mrbarge:
This looks good to merge as far as I'm concerned, but I'll let @sedroche get a chance to review and throw the lgtm/approved on there.

PR #26, opened 2020-06-25
Add cluster_upgrader builder

sedroche:


pkg/cluster_upgrader/cluster_upgrader.go
```diff
@@ -67,29 +65,38 @@ const (
 // Interface describing the functions of a cluster upgrader.
 //go:generate mockgen -destination=mocks/cluster_upgrader.go -package=mocks github.com/openshift/managed-upgrade-operator/pkg/cluster_upgrader ClusterUpgrader
 type ClusterUpgrader interface {
-	UpgradeCluster(c client.Client, m maintenance.Maintenance, upgradeConfig *upgradev1alpha1.UpgradeConfig, reqLogger logr.Logger) error
+	UpgradeCluster(upgradeConfig *upgradev1alpha1.UpgradeConfig, logger logr.Logger) error
+}
+
+//go:generate mockgen -destination=mocks/cluster_upgrader.go -package=mocks github.com/openshift/managed-upgrade-operator/pkg/cluster_upgrader ClusterUpgraderBuilder
+type ClusterUpgraderBuilder interface {
```
mrbarge, line 72: Has the mock been regenerated with this added interface? Not seeing it as part of the commit.

pkg/cluster_upgrader/cluster_upgrader.go
```diff
@@ -715,29 +736,29 @@ func ValidateUpgradeConfig(c client.Client, m maintenance.Maintenance, upgradeCo
...
 	if versions[0] != current {
-		log.Info(fmt.Sprintf("validation failed, current version %s is greater than desired %s", current, upgradeConfig.Spec.Desired.Version))
+		logger.Info(fmt.Sprintf("validation failed, current version %s is greater than desired %s", current, upgradeConfig.Spec.Desired.Version))
 		metrics.UpdateMetricValidationFailed(upgradeConfig.Name)
 		return false, fmt.Errorf("desired version %s is greater than current version %s", upgradeConfig.Spec.Desired.Version, current)
 	}
 
 	// Find the available versions from cincinnati
-	uuid, err := uuid.Parse(string(clusterVersion.Spec.ClusterID))
+	uuID, err := uuid.Parse(string(clusterVersion.Spec.ClusterID))
```
mrbarge, line 761: _Extremely_ minor and pedantic nit - would prefer this stay as uuid - it's not really a convention to have the mixed case for the term.

pkg/cluster_upgrader/cluster_upgrader.go
sedroche: I will regenerate once I start on fixing the tests. I wanted early feedback before I invest in the test :smile: 
Shall I commit them now so you can review?

pkg/cluster_upgrader/cluster_upgrader.go
sedroche: This var clashes with the package name and my IDE suggested to fix it. maybe `clusterId`?

pkg/cluster_upgrader/cluster_upgrader.go
mrbarge: Oh! No, that's fine, keep on going. :D  This all looks great to me.

pkg/cluster_upgrader/cluster_upgrader.go
mrbarge: clusterId sounds good :+1:

PR #27, opened 2020-06-30
Upgrade steps based on cluster state

sedroche:
Jira: https://issues.redhat.com/browse/OSD-3916

### DIFF
--- a/pkg/cluster_upgrader/cluster_upgrader.go
+++ b/pkg/cluster_upgrader/cluster_upgrader.go
@@ -13,9 +13,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/openshift/managed-upgrade-operator/pkg/maintenance"
-	"github.com/openshift/managed-upgrade-operator/pkg/metrics"
-
 	"github.com/blang/semver"
 	"github.com/go-logr/logr"
 	"github.com/google/uuid"
--- a/pkg/cluster_upgrader/cluster_upgrader.go
+++ b/pkg/cluster_upgrader/cluster_upgrader.go
@@ -567,8 +567,6 @@ func ControlPlaneUpgraded(c client.Client, metricsClient metrics.Metrics, m main
 	}
 	for _, c := range clusterVersion.Status.History {
 		if c.State == configv1.CompletedUpdate && c.Version == upgradeConfig.Spec.Desired.Version {
-			// send controlplane upgrade complete timestamp
-			metricsClient.UpdateMetricControlPlaneEndTime(time.Now(), upgradeConfig.Name, upgradeConfig.Spec.Desired.Version)
 			return true, nil
 		}
 	}
--- a/pkg/cluster_upgrader/cluster_upgrader.go
+++ b/pkg/cluster_upgrader/cluster_upgrader.go
@@ -2,10 +2,10 @@ package cluster_upgrader
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
-	"io/ioutil"
-	"net/http"
+	"github.com/openshift/managed-upgrade-operator/pkg/maintenance"
+	"github.com/openshift/managed-upgrade-operator/pkg/metrics"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"net/url"
 	"runtime"
 	"sort"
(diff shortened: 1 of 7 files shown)


sedroche:
I need to run a few more upgrades with this but initial runs were fine. Would appreciate any feedback

sedroche:
For metrics to work this needs to be run as a deployment. An image is here with these changes -> quay.io/sedroche/managed-upgrade-operator:idempotentish

pkg/maintenance/alertmanagerMaintenance.go
```diff
@@ -157,3 +186,14 @@ func createDefaultMatchers() []*amv2Models.Matcher {
 	inNamespaceAlertMatcher := createMatcher("namespace", "(^openshift.*|^kube.*|^redhat.*|^default$)", true)
 	return amv2Models.Matchers{nonCriticalAlertMatcher, inNamespaceAlertMatcher}
 }
+
+// Check if a silence with comment exists
+func silenceExists(mList *amSilence.GetSilencesOK, comment string) bool {
```
mrbarge, line 191: If an expired silence exists will it still return true here? does it also need to check payload.Status? 
(haven't confirmed if this is an issue, just wondering)

pkg/metrics/metrics.go
```diff
@@ -67,57 +118,180 @@ func init() {
...
-func (c *Counter) UpdateMetricClusterVerificationSucceeded(upgradeconfig string) {
+func (c *Counter) UpdateMetricClusterVerificationSucceeded(upgradeConfigName string) {
 	metricClusterVerificationFailed.With(prometheus.Labels{
-		nameLabel: upgradeconfig}).Set(
-			float64(0))
+		nameLabel: upgradeConfigName}).Set(
+		float64(0))
+}
+
+func getPromUrl(c client.Client) (*string, error) {
```
mrbarge, line 217: Minor nit.. when I think 'url' I think of the form with the protocol and path included, eg 'https://blah'.. maybe rename to getPromHost() to be clearer about what it's returning?

mrbarge:
I see where this is going and on the whole think it's a good change. I haven't tested it yet and want to do another pass over it now that I properly understand the goals, but so far it seems ok. Left some comments.

pkg/maintenance/alertmanagerMaintenance.go
sedroche: It does return expired silences. Each silence should only be created once and If it expires we get alerted for it.
If we check for active we can end up recreating a maintenance.

pkg/cluster_upgrader/cluster_upgrader.go
```diff
@@ -279,34 +300,47 @@ func EnsureExtraUpgradeWorkers(c client.Client, metricsClient metrics.Metrics, m
...
+	if isEqualVersion(cv, upgradeConfig) {
+		logger.Info(fmt.Sprintf("ClusterVersion is already set to Channel %s Version %s, skipping %s", desired.Channel, desired.Version, upgradev1alpha1.CommenceUpgrade))
+		return true, nil
+	}
+
+	if cv.Spec.DesiredUpdate != nil &&
+		cv.Spec.DesiredUpdate.Version == upgradeConfig.Spec.Desired.Version &&
+		cv.Spec.Channel == upgradeConfig.Spec.Desired.Channel {
 		return true, nil
 	}
```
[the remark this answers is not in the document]
sedroche: Nice. Thanks.

pkg/cluster_upgrader/cluster_upgrader.go
```diff
@@ -139,7 +140,17 @@ func Ordering() []upgradev1alpha1.UpgradeConditionType {
...
-	ok, err := performClusterHealthCheck(c, logger)
+	cv, err := getClusterVersion(c)
+	if err != nil {
+		return false, err
+	}
+	desired := upgradeConfig.Spec.Desired
+	if isEqualVersion(cv, upgradeConfig) {
+		logger.Info(fmt.Sprintf("ClusterVersion is already set to Channel %s Version %s, skipping %s", desired.Channel, desired.Version, upgradev1alpha1.UpgradePreHealthCheck))
+		return true, nil
+	}
```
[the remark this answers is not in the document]
sedroche: Sounds right. Not sure why I did it this way now.

pkg/metrics/metrics.go
sedroche: :+1:

sedroche:
@mrbarge Some updates have been pushed (can be seen in https://github.com/openshift/managed-upgrade-operator/compare/7fb0e9e73805587b5cfe1f3a4b89c2407203ca2d..37cd88ffdb33cfdb44bb1098faae17383e00b86c)

PR #29, opened 2020-07-01
Adding app-sre pipeline scripts

ravitri:
Adding app-sre pipeline scripts as per https://issues.redhat.com/browse/OSD-3807. Reference is with aws-account-operator..

### DIFF
--- a/hack/generate-operator-bundle.py
+++ b/hack/generate-operator-bundle.py
@@ -1,0 +1,141 @@
+#!/usr/bin/env python
+#
+# Generate an operator bundle for publishing to OLM. Copies appropriate files
+# into a directory, and composes the ClusterServiceVersion which needs bits and
+# pieces of our rbac and deployment files.
+#
+# Usage ./hack/generate-operator-bundle.py OUTPUT_DIR PREVIOUS_VERSION GIT_NUM_COMMITS GIT_HASH HIVE_IMAGE
+#
+# Commit count can be obtained with: git rev-list 9c56c62c6d0180c27e1cc9cf195f4bbfd7a617dd..HEAD --count
+# This is the first hive commit, if we tag a release we can then switch to using that tag and bump the base version.
+
+import datetime
+import os
+import sys
+import yaml
+import shutil
+
+# This script will append the current number of commits given as an arg
+# (presumably since some past base tag), and the git hash arg for a final
+# version like: 0.1.189-3f73a592
+VERSION_BASE = "0.2"
+
+OPERATOR_NAME="managed-upgrade-operator"
+
+if len(sys.argv) != 6:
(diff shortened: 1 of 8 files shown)


hack/generate-operator-bundle.py
```diff
@@ -0,0 +1,141 @@
...
+import datetime
+import os
+import sys
+import yaml
+import shutil
+
+# This script will append the current number of commits given as an arg
+# (presumably since some past base tag), and the git hash arg for a final
+# version like: 0.1.189-3f73a592
+VERSION_BASE = "0.2"
```
mrbarge, line 21: Maybe we can start at 0.1

hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -0,0 +1,41 @@
...
+objects:
+- apiVersion: operators.coreos.com/v1alpha1
+  kind: CatalogSource
+  metadata:
+    name: managed-upgrade-operator-catalog
+  spec:
+    sourceType: grpc
+    image: ${REGISTRY_IMG}:${CHANNEL}-${IMAGE_TAG}
+    displayName: managed-upgrade-operator Registry
+    publisher: SRE
```
mrbarge, line 23: We should maybe be setting an affinity for infra nodes here, as suggested by OSD-4113. See cloud-ingress-operator as an example.

sedroche:
Looks good to me but I'm not hugely familiar with the process. I would appreciate a quick run through if you have time

hack/generate-operator-bundle.py
ravitri: Ahhh!! I swear I had changed this. Wonder why it was not changed.. Thanks to catch it though.. Will fix this.

hack/olm-registry/olm-artifacts-template.yaml
ravitri: Nice point Matt.. As per https://github.com/openshift/cloud-ingress-operator/commit/227c095f11de23183bcdc60d21345a296f0e19e0#diff-e12c68acf5db1ed8c169f6a9c8a722a6 I see that requiredDuringSchedulingIgnoredDuringExecution is used to schedule operator pod on infra node. Whereas, https://github.com/openshift/rbac-permissions-operator/pull/67/files#diff-5e5f804a8ee2fb8c81711b728eb5a21d is still a discussion where its suggested to use preferredDuringSchedulingIgnoredDuringExecution. I think its better to have a separate JIRA card for this and address it after this is decided. Sounds ok?

ravitri:
@sedroche - Sure.. to be honest, I myself am struggling to understand it all because of so many pieces involved but I can have a session for it for team to have an overview.

hack/olm-registry/olm-artifacts-template.yaml
mrbarge: Sure, a separate card sounds fine to me, let's get clarification on required vs preferred and then implement.

PR #30, opened 2020-07-02
verify the certs for the prometheus route in the tls accessing

bmeng:
Remove the insecure access for the prometheus url for OSD-3945


sedroche:
:+1:

PR #32, opened 2020-07-02
Fixing standard.mk to have proper image tagging for push

ravitri:
As per https://ci.int.devshift.net/job/openshift-managed-upgrade-operator-gh-build-master/4/console, the below skopeo command failed since it didn't find any managed-upgrade-operator:latest image at all which should be present locally before tagging. This fix will create the required image and tag and push to quay repo.

`skopeo copy --dest-creds ****:**** docker-daemon:managed-upgrade-operator:latest docker://quay.io/app-sre/managed-upgrade-operator:latest`

The IMG variable is passed through https://github.com/openshift/managed-upgrade-operator/blob/master/hack/app_sre_build_deploy.sh#L18 for tagging.


standard.mk
```diff
@@ -20,7 +20,8 @@ COMMIT_NUMBER=$(shell git rev-list `git rev-list --parents HEAD | egrep "^[a-f0-
 CURRENT_COMMIT=$(shell git rev-parse --short=7 HEAD)
 OPERATOR_VERSION=$(VERSION_MAJOR).$(VERSION_MINOR).$(COMMIT_NUMBER)-$(CURRENT_COMMIT)
 
-OPERATOR_IMAGE_URI=$(IMAGE_REGISTRY)/$(IMAGE_REPOSITORY)/$(IMAGE_NAME):v$(OPERATOR_VERSION)
+IMG?=$(IMAGE_REGISTRY)/$(IMAGE_REPOSITORY)/$(IMAGE_NAME):v$(OPERATOR_VERSION)
```
dofinn, line 23: is the `v` in `:v$` a typo?

ravitri: Thanks a lot to catch that! This wasn't a typo. Actually, I checked working logs for aws-account-operator https://ci-int-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/osd-operators/job/openshift-aws-account-operator-gh-build-master/289/consoleFull and also https://github.com/openshift/aws-account-operator/blob/master/standard.mk#L23. The 'v' is configured as part of this variable but I don't see this used while actual tagging. The IMG variable is as such passed with "managed-upgrade-operator:latest" to `docker build`. The tags for aws-account-operator in https://quay.io/repository/app-sre/aws-account-operator?tab=tags are based on GIT_HASH variable from what I understand. For now, have removed 'v' since the working logs didn't refer it. We can add it in future if required.

PR #33, opened 2020-07-03
Masters are reported upgraded by CVO

sedroche:
CVO reports if master nodes are upgraded so there is no need for the `AllMastersUpgraded` check.
And depending on how long the control plane lasts, all workers may already be upgraded.


sedroche:
/hold

as its not tested yet

mrbarge:
Tested it with upgrades under varying conditions and it worked fine.

/lgtm

PR #37, opened 2020-07-07
BDD tests for maintenance

dofinn:
```
Running Suite: Maintenance Suite
================================
Random Seed: 1594099769
Will run 7 of 7 specs

•••••••
Ran 7 of 7 Specs in 0.000 seconds
SUCCESS! -- 7 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS
coverage: 70.4% of statements

Ginkgo ran 1 suite in 2.453542127s
Test Suite Passed
```


pkg/maintenance/maintenance_test.go
```diff
@@ -0,0 +1,169 @@
...
+		testComment           = "test comment"
+		testOperatorName      = "managed-upgrade-operator"
+		testCreatedByOperator = testOperatorName
+		testCreatedByTest     = "Tester the Creator"
+		testNow               = strfmt.DateTime(time.Now().UTC())
+		testEnd               = strfmt.DateTime(time.Now().UTC().Add(90 * time.Minute))
+		testVersion           = "V-1.million.25"
+		testGettableSilences  amv2Models.GettableSilences
+
+		// Create test silence NOT created by the operator
```
mrbarge, line 35: Minor nit, this comment seems to indicate the opposite of what's being created.

mrbarge:
Other than one trivial noted thing, this looks great to me. I'll let more :eyes: have an opportunity to see it before lgtm'ing it.

pkg/maintenance/alertManagerSilenceClient.go
```diff
@@ -9,12 +9,19 @@ import (
 	"net/http"
 )
 
+//go:generate mockgen -destination=alertManagerSilenceMock.go -package=maintenance github.com/openshift/managed-upgrade-operator/pkg/maintenance AlertManagerSilencer
```
mrbarge, line 12: If I recall, was the reason for not putting it the 'mocks' package dir due to the cyclic import that results in?

dofinn: correct

pkg/maintenance/maintenance_test.go
dofinn: So my comment is incorrect. This was a validation that the operator will only retrieve maintenance's created by the operator. The actual test got moved to another spot in the file. Ill ammend

PR #38, opened 2020-07-07
Changing saas repo to app-interface for catalog creation

ravitri:
Changing the saas repo URL to app-interface.

This is based upon https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/6053.

### DIFF
--- a/hack/app_sre_create_image_catalog.sh
+++ b/hack/app_sre_create_image_catalog.sh
@@ -27,8 +27,8 @@ git clone \
 REMOVED_VERSIONS=""
 if [[ "$REMOVE_UNDEPLOYED" == true ]]; then
     DEPLOYED_HASH=$(
-        curl -s "https://gitlab.cee.redhat.com/service/saas-osd-operators/raw/master/${_OPERATOR_NAME}-services/${_OPERATOR_NAME}.yaml" | \
-            docker run --rm -i evns/yq -r '.services[]|select(.name="${_OPERATOR_NAME}").hash'
+        curl -s "https://gitlab.cee.redhat.com/service/app-interface/raw/master/data/services/osd-operators/cicd/saas/saas-${_OPERATOR_NAME}.yaml" | \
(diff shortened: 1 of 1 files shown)


hack/app_sre_create_image_catalog.sh
```diff
@@ -27,8 +27,8 @@ git clone \
 REMOVED_VERSIONS=""
 if [[ "$REMOVE_UNDEPLOYED" == true ]]; then
     DEPLOYED_HASH=$(
-        curl -s "https://gitlab.cee.redhat.com/service/saas-osd-operators/raw/master/${_OPERATOR_NAME}-services/${_OPERATOR_NAME}.yaml" | \
-            docker run --rm -i evns/yq -r '.services[]|select(.name="${_OPERATOR_NAME}").hash'
+        curl -s "https://gitlab.cee.redhat.com/service/app-interface/raw/master/data/services/osd-operators/cicd/saas/saas-${_OPERATOR_NAME}.yaml" | \
+            docker run --rm -i evns/yq -r '.resourceTemplates[]|select(.name="managed-upgrade-operator").targets[]|select(.namespace["$ref"]=="/services/osd-operators/namespaces/managed-upgrade-operator-production.yml")|.ref'
```
bmeng, line 31: it is still able to use variable if using double quotas with escape ?

ravitri: Sorry, which variable is escaped? I don't see it

mrbarge: I think @bmeng is suggesting the 'docker' line could still use the OPERATOR_NAME variable if it used double-quotes for the argument and escaped the inner-doublequotes. eg

`docker run --rm -i evns/yq -r ".resourceTemplates[]|select(.name=\"${_OPERATOR_NAME}\").targets[]|select(.namespace[\"\$ref\"]==\"/services/osd-operators/namespaces/${_OPERATOR_NAME}-production.yml\")|.ref"`

Or alternatively maybe something like:

`docker run --rm -i evns/yq -r '.resourceTemplates[]|select(.name="'"${_OPERATOR_NAME}"'").targets[]|select(.namespace["$ref"]=="/services/osd-operators/namespaces/'"${_OPERATOR_NAME}"'-production.yml")|.ref'`

PR #40, opened 2020-07-08
OSD-4334 and Scaler interface

sedroche:
This contains a fix for https://issues.redhat.com/browse/OSD-4334 -> https://github.com/openshift/managed-upgrade-operator/commit/861f4f640ec77e267c7cfdafa477d2f18759ab42

But also includes:
- New scaler pkg and interface
- cluster upgrader scaler tests moved to scaler pkg


pkg/cluster_upgrader/cluster_upgrader.go
```diff
@@ -569,6 +436,7 @@ func ControlPlaneUpgraded(c client.Client, metricsClient metrics.Metrics, m main
 	}
 
 	return false, nil
+
```
wanghaoran1988, line 439: remove this

pkg/cluster_upgrader/cluster_upgrader_test.go
```diff
@@ -3,6 +3,7 @@ package cluster_upgrader
 import (
 	"context"
 	"fmt"
+	"github.com/openshift/managed-upgrade-operator/pkg/scaler"
```
wanghaoran1988, line 6: put this in bottom line ? we should better group the imports

sedroche: I just rely on `go fmt` for ordering. It keeps it consistent. What order do you suggest?

wanghaoran1988: Split the imports like this with an empty line : https://github.com/openshift/cluster-version-operator/blob/master/pkg/cvo/availableupdates.go#L4-L18 ?

sedroche: Can do. `go fmt` seems to respect the groupings

PR #44, opened 2020-07-09
Config for upgrader

sedroche:
Jira: https://issues.redhat.com/browse/OSD-4024

External configuration in configmap


wanghaoran1988:
We need have a default configmap created after the operator is installed and managed by OLM?

sedroche:
> We need have a default configmap created after the operator is installed and managed by OLM?

@wanghaoran1988 The intention (I believe) is to have this as part of the `SelectorSyncSet`
@mrbarge Do we need to incorporate this into your generation script?

mrbarge:
> @mrbarge Do we need to incorporate a configmap into your generation script if this lands?

We could possibly do that. Or, because this value is likely to change less often, we could put this and any other configuration items we invent in a separate ConfigMap SelectorSyncSet that's managed in managed-cluster-config. That'd be my preferred option personally but either way a SSS seems like the way to go.

sedroche:
If we can have a default set by OLM I think we should too

mrbarge:
Seems fine to me so far. You may already be adding this, but it'll be good to get some data validation and tests happening around the configmap, much like what the GCP Project Operator [is doing](https://github.com/openshift/gcp-project-operator/tree/master/pkg/configmap). For instance, preventing negative values, words, etc.  (I tried both of those - the former is rejected by AM because a Silence can't be in the past, and the second just causes the controller to barf-loop when unmarshalling).

sedroche:
@wanghaoran1988 I don't think CSVs can contain other resources.
We can create a default one on startup instead?

mrbarge:
Tested fine and PR looks good:
- Verified behaviour of operator when configmap is missing. (it errors, but gracefully)
- Verified setting control plane window to 15m, observed a 15m silence window created.
- Verified setting upgrade window timeout to 10m, observed window breach when upgrading 10+m afterwards.
Code seems fine, I think this is good to merge.

pkg/osd_cluster_upgrader/config.go
```diff
@@ -0,0 +1,42 @@
...
+type nodeDrain struct {
+	TimeOut time.Duration `yaml:"timeOut"`
+}
+
+func (cfg *osdUpgradeConfig) IsValid() error {
+	if cfg.Maintenance.ControlPlaneTime <= 0 {
+		return fmt.Errorf("Config maintenace controlPlaneTime out is invalid")
+	}
+	if cfg.Maintenance.WorkerNodeTime <= 0 {
+		return fmt.Errorf("Config maintenace workerNddeTime is invalid")
```
mrbarge, line 32:
```suggestion
		return fmt.Errorf("Config maintenace workerNodeTime is invalid")
```

PR #46, opened 2020-07-10
Alert on scaling up

sedroche:
https://issues.redhat.com/browse/OSD-4409

Setting scale metric on timeout (alerting still to be done in https://issues.redhat.com/browse/OSD-4382)


mrbarge:
I ran this through a real test* and it behaved as expected. 

*I customized the timeout to be 1 minute, and then just kept on deleting the scaled machine until it re-reconciled and triggered the error. Verified the metric was set:
`upgradeoperator_scaling_failed{upgradeconfig_name="example-upgrade-config"} 1`
Then verified that - after scaling was allowed to properly finish - it was reset:
`upgradeoperator_scaling_failed{upgradeconfig_name="example-upgrade-config"} 0`

/lgtm

PR #52, opened 2020-07-14
Add ClusterUpgrader Type

sedroche:
Jira: https://issues.redhat.com/browse/OSD-4472

Adding a type to the CRD to indicate the type of cluster upgrader we want. (Some specializations may be needed such as OSD-force or ARO) but this allows type to be configurable.


deploy/crds/upgrade.managed.openshift.io_v1alpha1_upgradeconfig_cr.yaml
```diff
@@ -3,6 +3,7 @@ kind: UpgradeConfig
 metadata:
   name: example-upgrade-config
 spec:
+  type: "OSD"
   desired:
     channel: "fast-4.4"
     force: false
```
mrbarge, line 9: Sorry to be a pain, should have caught this in Haoran's PR. Can we add the upgradeAt and proceed fields here. (upgradeAt is required for it to actually load)
```suggestion
  type: "OSD"
  upgradeAt: "2020-01-01T00:00:00Z"
  proceed: true
  desired:
    channel: "fast-4.4"
    force: false
```

PR #59, opened 2020-07-16
Implement upgradeAt logic

wanghaoran1988:


pkg/osd_cluster_upgrader/upgrader.go
```diff
@@ -608,7 +608,7 @@ func performValidateUpgradeConfig(c client.Client, upgradeConfig *upgradev1alpha
 	sort.Strings(versions)
 	if versions[0] != current {
 		logger.Info(fmt.Sprintf("validation failed, current version %s is greater than desired %s", current, upgradeConfig.Spec.Desired.Version))
-		return false, fmt.Errorf("desired version %s is greater than current version %s", upgradeConfig.Spec.Desired.Version, current)
+		return false, fmt.Errorf("desired version %s is small than current version %s", upgradeConfig.Spec.Desired.Version, current)
```
mrbarge, line 611:
```suggestion
		return false, fmt.Errorf("desired version %s is older than current version %s", upgradeConfig.Spec.Desired.Version, current)
```

pkg/controller/upgradeconfig/upgradeconfig_controller.go
```diff
@@ -157,5 +157,18 @@ func (r *ReconcileUpgradeConfig) Reconcile(request reconcile.Request) (reconcile
...
+		log.Info("upgrade cannot be proceed", "proceed", upgradeConfig.Spec.Proceed)
+		return false
+	}
+	upgradeTime, err := time.Parse(time.RFC3339, upgradeConfig.Spec.UpgradeAt)
+	if err != nil {
+		log.Error(err, "failed to parse spec.upgradeAt", upgradeConfig.Spec.UpgradeAt)
+		return false
+	}
+	now := time.Now()
+	if now.After(upgradeTime) && upgradeTime.Add(15*time.Minute).After(now) {
```
dofinn, line 170: This should be an or statement?

mrbarge: I had to scratch my head a bit at the logic too, but it checks out (assuming it's the behaviour we want).
It's basically saying "If the current time is after the scheduled time, and is also 15 minutes after the upgrade time, don't upgrade". ie. if the operator is more than 15 minutes late getting around to the request, we're not going to upgrade.
It's the same thing I was getting at on Slack really. I feel a bit uneasy putting that policy in code though. But we do need to clearly define what the operator's behaviour should be - ie. how "late" is too late when commencing an upgrade?

PR #66, opened 2020-07-20
update the metrics to record the upgrade started

bmeng:
Update the starttime stamp to the started marker. Since the start timestamp will be cleared with the pod restarted.
Move the update metric call to the earlier of the func CommenceUpgrade to make sure it will be set after the pod recreated.

The will be conflict with @mrbarge 's PR. Put it here for review firstly, and will rebase later.


bmeng:
https://github.com/openshift/managed-cluster-config/pull/395
Promethues rules PR

bmeng:
Close this one in favor of #68

PR #68, opened 2020-07-20
implement the timeout func into operator

bmeng:
https://issues.redhat.com/browse/OSD-4508

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -4,16 +4,17 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	routev1 "github.com/openshift/api/route/v1"
-	"github.com/prometheus/client_golang/prometheus"
 	"io/ioutil"
-	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/types"
 	"net/http"
-	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/metrics"
 	"strings"
 	"time"
+
+	routev1 "github.com/openshift/api/route/v1"
+	"github.com/prometheus/client_golang/prometheus"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
 )
 
 const (
@@ -37,14 +38,17 @@ type Metrics interface {
 	UpdateMetricClusterVerificationSucceeded(string)
 	UpdateMetricUpgradeWindowNotBreached(string)
 	UpdateMetricUpgradeWindowBreached(string)
+	UpdateMetricUpgradeControlPlaneTimeout(string, string)
+	UpdateMetricUpgradeWorkerTimeout(string, string)
 	IsMetricUpgradeStartTimeSet(upgradeConfigName string, version string) (bool, error)
 	IsMetricControlPlaneEndTimeSet(upgradeConfigName string, version string) (bool, error)
 	IsMetricNodeUpgradeEndTimeSet(upgradeConfigName string, version string) (bool, error)
+	IsSilenceActive() (bool, error)
 	Query(query string) (*AlertResponse, error)
 }
 
 //go:generate mockgen -destination=mocks/metrics_builder.go -package=mocks github.com/openshift/managed-upgrade-operator/pkg/metrics MetricsBuilder
-type MetricsBuilder interface{
+type MetricsBuilder interface {
 	NewClient(c client.Client) (Metrics, error)
 }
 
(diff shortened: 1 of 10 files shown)


pkg/metrics/metrics.go
```diff
@@ -133,6 +137,16 @@ var (
 		Name:      "upgrade_window_breached",
 		Help:      "Failed to commence upgrade during the upgrade window",
 	}, []string{nameLabel})
+	metricUpgradeControlPlaneTimeout = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Subsystem: metricsTag,
+		Name:      "controlplane_timeout",
```
wanghaoran1988, line 142: This name doesn't sounds good to me, we probably should revisit the existing metrics names to have a better pattern like "managed_upgrade_xxx"

bmeng: the metricsTag is the prefix for all the metrics.
So it will be `upgradeoperator_xxx` now. If the `upgradeoperator` does not make sense.
We can change it in const

pkg/osd_cluster_upgrader/upgrader.go
```diff
@@ -247,6 +248,27 @@ func AllWorkersUpgraded(c client.Client, scaler scaler.Scaler, metricsClient met
...
+		if uc.Version == upgradeConfig.Spec.Desired.Version {
+			for _, ucc := range uc.Conditions {
+				if ucc.Type == "AllWorkerNodesUpgraded" {
+					nodeCompleteTime = ucc.CompleteTime
+				}
+			}
+		}
+	}
+
+	if !silenceActive && nodeCompleteTime == nil {
```
wanghaoran1988, line 267: Can we do node upgrade timeout per node ?

pkg/metrics/metrics.go
wanghaoran1988: thanks, missed that

pkg/osd_cluster_upgrader/upgrader.go
bmeng: We already had the logic to setup the silence based on the node number.
As discussed with john, it is more easier to use the existing silence.

bmeng: per node? Do you mean check the timeout for every node separately?

wanghaoran1988: @bmeng yes, I know it's not easy, but we should have that, like if we have 10 nodes, and upgrade node stuck at first one, we should got alert if first node upgrade timeout instead of 10 nodes timeout, WDYT?

bmeng: Yeah, it is not quite easy IMO. Seems like the nodes will be upgraded parallel? 

Another thing is, we setup the silence for the warning and lower severity. 
As the condition you mentioned, if the node stuck, I think some of the operators will be Degraded. 
So we will get paged regardless of the upgrade, but the cluster healthy.

pkg/osd_cluster_upgrader/upgrader.go
```diff
@@ -407,6 +429,25 @@ func ControlPlaneUpgraded(c client.Client, scaler scaler.Scaler, metricsClient m
 			isCompleted = true
 		}
 	}
+
+	var upgradeStartTime, controlPlaneCompleteTime *metav1.Time
+	for _, uc := range upgradeConfig.Status.History {
```
sedroche, line 434: Can we base this on ClusterVersion startedTime for version? Seems more correct if we our timeout is for controlplane

pkg/osd_cluster_upgrader/upgrader.go
```diff
@@ -247,6 +248,27 @@ func AllWorkersUpgraded(c client.Client, scaler scaler.Scaler, metricsClient met
 		return false, err
 	}
 
+	silenceActive, err := metricsClient.IsSilenceActive()
+	if err != nil {
+		return false, err
+	}
+
+	var nodeCompleteTime *metav1.Time
```
sedroche, line 256: If we get past `ok, err := nodesUpgraded(c, "worker", logger)` we always have a nodeCompleteTime I think.
So is the check just `if !silenceActive {`?

pkg/osd_cluster_upgrader/upgrader.go
sedroche: I think we will eventually need to refine this watch per node but for M1 this is ok

sedroche: I think we will eventually need to refine this watch per node but for M1 this is ok

wanghaoran1988: Sure, @bmeng Could you create one ticket to track this ?

bmeng:
@sedroche updated the PR a bit, can you help review again?

pkg/metrics/metrics.go
```diff
@@ -235,6 +265,19 @@ func (c *Counter) IsMetricNodeUpgradeEndTimeSet(upgradeConfigName string, versio
 	return false, nil
 }
 
+func (c *Counter) IsSilenceActive() (bool, error) {
```
sedroche, line 268: Could you also check that the silence is created by the managed-upgrade-operator?
See -> https://github.com/openshift/managed-upgrade-operator/blob/ff4fbb232623e0609b10f748adc1bb73fcc0891c/pkg/maintenance/alertmanagerMaintenance.go#L166

PR #69, opened 2020-07-20
OSD-4548: Removing defaults from CRD as not supported in v1beta1

ravitri:
After reverting the CRD version from v1 to v1beta1, the creation of CRD failed with below error:
```
$ oc create -f https://raw.githubusercontent.com/openshift/managed-upgrade-operator/master/deploy/crds/upgrade.managed.openshift.io_upgradeconfigs_crd.yaml
The CustomResourceDefinition "upgradeconfigs.upgrade.managed.openshift.io" is invalid: 
* spec.validation.openAPIV3Schema.properties[status].properties[history].items.properties[phase].default: Forbidden: must not be set
* spec.validation.openAPIV3Schema.properties[spec].properties[proceed].default: Forbidden: must not be set
* spec.validation.openAPIV3Schema.properties[spec].properties[PDBForceDrainTimeout].default: Forbidden: must not be set
* spec.validation.openAPIV3Schema.properties[spec].properties[desired].properties[force].default: Forbidden: must not be set
```
This PR is to remove the "defaults" from CRD as it is not supported as part of v1beta1 version of CRD. Seems this CRD version change wasn't handled by operator-sdk. 

Checked creation of CRD and it works fine as expected on OCP 4.4. This PR is part of https://issues.redhat.com/browse/OSD-4548.

### DIFF
--- a/deploy/crds/upgrade.managed.openshift.io_upgradeconfigs_crd.yaml
+++ b/deploy/crds/upgrade.managed.openshift.io_upgradeconfigs_crd.yaml
@@ -54,7 +54,6 @@ spec:
             and upgrade window and freeze window
           properties:
             PDBForceDrainTimeout:
-              default: 60
               description: The maximum grace period granted to a node whose drain
                 is blocked by a Pod Disruption Budget, before that drain is forced.
                 Measured in minutes.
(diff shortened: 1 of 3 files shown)


mrbarge:
The olm template will also need updating.

ravitri:
@mrbarge - Oh yes! Thanks for catching it.. Missed it this time.. Removed it too now

PR #71, opened 2020-07-23
OSD-4925 validating UpgradeConfig prior to upgrade

dofinn:
This PR:
* Moves validation into its own packages
* Uses a validation builder/interface as apart of the reconciler
* refactors validation bar cincinatti verification
* Creates a 'util' package for testing

Extra:
* Returns reconciler error if status == ""
* Some house keeping comments in types

We should follow this card with test refactoring and standardisation.


pkg/controller/upgradeconfig/upgradeconfig_controller.go
```diff
@@ -117,19 +122,56 @@ func (r *ReconcileUpgradeConfig) Reconcile(request reconcile.Request) (reconcile
...
 
 	switch status {
-	case "", upgradev1alpha1.UpgradePhaseNew, upgradev1alpha1.UpgradePhasePending:
-		// TODO verify if it's time to do upgrade, if no, set to "pending", if it's yes, perform upgrade, and set status to "upgrading"
-		reqLogger.Info("checking whether it's ready to do upgrade")
+	case upgradev1alpha1.UpgradePhaseNew, upgradev1alpha1.UpgradePhasePending:
+		reqLogger.Info("Validating UpgradeConfig")
+
+		// Get current ClusterVersion
+		cv, err := osd_cluster_upgrader.GetClusterVersion(r.client)
```
mrbarge, line 138: A minor nit that doesn't necessarily need addressing here, but maybe we eventually need to refactor the code used by GetClusterVersion (and the GetCurrentVersion used by the validator) into something outside of the OSD Cluster Upgrader. Mainly because the controller and the validator are pretty OCP-agnostic things that now have a dependency on code sitting in the OSD-specific package, which gives me a few irritations design-wise.

pkg/validation/validation.go
```diff
@@ -0,0 +1,138 @@
...
+type ValidationBuilder interface {
+	NewClient(client client.Client) (Validator, error)
+}
+
+// validationBuilder is an empty struct that enables instantiation of this type and its
+// implimented interface.
+type validationBuilder struct{}
+
+// NewClient returns a Validator interface or an error if one occurs.
+func (vb *validationBuilder) NewClient(client client.Client) (Validator, error) {
```
mrbarge, line 136: The constructor doesn't seem to be saving the kube client to the struct or using it, is it needed?
Could either remove the client dependency (unless you foresee a need for it in future revisions), or alternately keep it and store it and also maybe move the GetClusterVersion() call (which uses the client) out of the controller and into the IsValidUpgradeConfig() function.

pkg/validation/validation.go
```diff
@@ -0,0 +1,138 @@
...
+	case -1:
+		logger.Info(fmt.Sprintf("%s is less then %s", dV, cV))
+		logger.Info(fmt.Sprintf("Downgrading cluster is not supported. Not Proceeding to %s", dV))
+		return false
+	case 0:
+		logger.Info(fmt.Sprintf("%s is equal to %s", dV, cV))
+		logger.Info(fmt.Sprintf("Cluster is already on version %s", cV))
+		return false
+	case 1:
+		logger.Info(fmt.Sprintf("%s is greater then %s", dV, cV))
```
mrbarge, line 120: Can we add a case for "default" that returns false, because otherwise if semver's Compare() goes wonky and returns a "2" for whatever reason it'll fall through to returning true.

pkg/validation/validation.go
```diff
@@ -0,0 +1,138 @@
...
+	// Check whether the desired version exists in availableUpdates
+	found := false
+	for _, v := range cvoUpdates {
+		if v.Version == dv && !v.Force {
+			found = true
+		}
+	}
+
+	if !found {
+		logger.Info(fmt.Sprintf("Failed to find the desired version %s in channel %s", desiredVersion, desiredChannel))
```
ravitri, line 98: As per https://github.com/openshift/cluster-version-operator/blob/master/pkg/cincinnati/cincinnati.go#L120 and https://github.com/openshift/cluster-version-operator/blob/master/lib/validation/validation.go there are certain validations done already by CVO. For this check regarding version check from channel, do we still need it again as part of upgrade operator? I think it would be also helpful to explore the re-sync period when CVO would perform the validation again after a change and let CVO report error if something is wrong. This is about keeping the validation checks of product via product itself if possible.

pkg/controller/upgradeconfig/upgradeconfig_controller_test.go
```diff
@@ -192,12 +241,32 @@ var _ = Describe("UpgradeConfigController", func() {
 
 				Context("When the cluster is ready to upgrade", func() {
 					Context("When a cluster upgrade client can't be built", func() {
+						var clusterVersionList *configv1.ClusterVersionList
+						BeforeEach(func() {
+							clusterVersionList = &configv1.ClusterVersionList{
```
mrbarge, line 246: Minor nit, looks like the `clusterVersionlist` variable could all be moved to one of the outer (even the outermost) BeforeEach() contexts, that would tidy things up a bit..

pkg/validation/validation.go
```diff
@@ -0,0 +1,138 @@
...
+		return false, err
+	}
+	upstreamURI, err := url.Parse(string(cV.Spec.Upstream))
+	if err != nil {
+		return false, err
+	}
+
+	updates, err := cincinnati.NewClient(clusterId, nil, nil).GetUpdates(upstreamURI, runtime.GOARCH, desiredChannel, currentVersion)
+	if err != nil {
+		return false, err
```
mrbarge, line 78: I mentally go back and forth on the Cincinnati stuff and don't have a horse in the race for whether we call it or not, but outside of that:

- If CVO is already reporting the version as valid, do we need to do a Cincinnati check at all? This could help remove Cinny as a blocking dependency in the majority of situations.
- If CVO is not reporting the version as valid, but Cincinnati does, do we still go ahead? (ie. is this a case where urgent CVE must go out, CVO hasn't synced yet, but we know that Cincinnati thinks everything's a-ok)
- If Cincinnati is down, could the error message please indicate that - we are currently relying on the error returned by the cincinnati client to contain useful data, which it may not. so eg
```suggestion
		return false, fmt.Errorf("Failed to pull Cincinnati updates: %v", err)
```

mrbarge:
I'm done with a review personally, outside of what I've noted in comments above things all look good.

pkg/validation/validation.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/go-logr/logr"
+)
+
+// NewBuilder returns a validationBuilder object that implements the ValidationBuilder interface.
+func NewBuilder() ValidationBuilder {
+	return &validationBuilder{}
+}
+
+// Validator knows how to validate UpgradeConfig CRs.
+//go:generate mockgen -destination=mockValidation.go -package=validation github.com/openshift/managed-upgrade-operator/pkg/validation Validator
```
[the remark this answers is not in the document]
dofinn: sure. I did have this there at one stage. I think make squealed because my repo was not clean when I made the change. Resolved: f351af304bf7c65ef25f052469b86f6f430201c5

pkg/apis/upgrade/v1alpha1/upgradeconfig_types.go
```diff
@@ -150,7 +160,7 @@ type Update struct {
 	// Version of openshift release
 	// +kubebuilder:validation:Type=string
 	Version string `json:"version"`
-	// Channel we gonna use for upgrades
+	// Channel used for upgrades
```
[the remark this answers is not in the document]
dofinn: Good catch, I like it: df885c6c025b7db2de2ea5fe7b11d3c6e4db9987

pkg/controller/upgradeconfig/upgradeconfig_controller_test.go
dofinn: Yeah I think that kind of work can be done in a refactor? I would presume we are duplicating vars else where in this file.

pkg/validation/validation.go
dofinn: i think we leave this in until OCM is in front of the service and is validating against CIS.

pkg/validation/validation.go
dofinn: so AFAIK, we were leaving cincinatti validation in until OCM was infront of the service and sending down the curated versions in the upgradeConfig CR. If we want to change cincinatti validation, we should define our requirements in another task.

pkg/controller/upgradeconfig/upgradeconfig_controller_test.go
mrbarge: Yeah, the overall refactor will be a good place to just take a fresh look at things and re-structure to a common approach. Agree - don't need to do anything on this one here.  Refs OSD-4590

pkg/validation/validation.go
dofinn: good catch: 9e82bc63580cd51064f7a8e14ac3d3deca8db53e

pkg/validation/validation.go
dofinn: Resolved dep: ad2e4705621ded503b2b5a67cd942055fff39b2c

pkg/controller/upgradeconfig/upgradeconfig_controller.go
sedroche: +1 can we add to https://issues.redhat.com/browse/OSD-4168?

pkg/controller/upgradeconfig/upgradeconfig_controller.go
```diff
@@ -117,19 +122,56 @@ func (r *ReconcileUpgradeConfig) Reconcile(request reconcile.Request) (reconcile
...
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
+		// Build a Validator
```
sedroche, line 143: nitpick but the comments are unecessary

pkg/controller/upgradeconfig/upgradeconfig_controller.go
```diff
@@ -117,19 +122,56 @@ func (r *ReconcileUpgradeConfig) Reconcile(request reconcile.Request) (reconcile
 	}
 
 	status := history.Phase
-	reqLogger.Info("current cluster status", "status", status)
+
+	// status should never be an empty string at this stage. Explicity test and return error if it is.
+	if status == "" {
```
sedroche, line 127: Is this necessary? Cant see how it will be empty

dofinn: https://github.com/openshift/managed-upgrade-operator/pull/71/files/ad2e4705621ded503b2b5a67cd942055fff39b2c#diff-416322dadc9a204248e5e54f38904009L123

We currently accept it as valid. Id prefer to handle this explicitly.

sedroche: I'd remove that too :joy:

pkg/validation/validation.go
ravitri: As discussed on Slack, this could be taken up in https://issues.redhat.com/browse/OSD-4152 further on after considering how OCM decides to handle it.. For now to keep it as it is..

mrbarge:
Branch has tested successfully on a real cluster, @sedroche I'll let you have the honours of slapping an lgtm on it if/when you're happy.

pkg/controller/upgradeconfig/upgradeconfig_controller.go
dofinn: added

PR #72, opened 2020-07-26
Move history to reconciler and remove extra conditions

sedroche:
https://issues.redhat.com/browse/OSD-4497

Fixes a few things.

- ClusterUpgraders are now plugable components so to keep the history status updating consistent I have moved it to the reconciler. ClusterUpgraders will report their status and condition.
- We only display the one condition in our printer columns so moving to just saving the current one (Status gets very unreadable after a few upgrades) Condition reports the "the latest available observations of an object’s state"
- Test updates

### DIFF
--- a/pkg/controller/upgradeconfig/upgradeconfig_controller.go
+++ b/pkg/controller/upgradeconfig/upgradeconfig_controller.go
@@ -4,7 +4,10 @@ import (
 	"context"
 	"time"
 
+	"github.com/go-logr/logr"
+	"github.com/hashicorp/go-multierror"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -15,7 +18,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	upgradev1alpha1 "github.com/openshift/managed-upgrade-operator/pkg/apis/upgrade/v1alpha1"
-	"github.com/openshift/managed-upgrade-operator/pkg/cluster_upgrader_builder"
+	cub "github.com/openshift/managed-upgrade-operator/pkg/cluster_upgrader_builder"
 	"github.com/openshift/managed-upgrade-operator/pkg/metrics"
 	"github.com/openshift/managed-upgrade-operator/pkg/osd_cluster_upgrader"
 	"github.com/openshift/managed-upgrade-operator/pkg/validation"
@@ -37,7 +40,7 @@ func newReconciler(mgr manager.Manager) reconcile.Reconciler {
 		client:                 mgr.GetClient(),
 		scheme:                 mgr.GetScheme(),
 		metricsClientBuilder:   metrics.NewBuilder(),
-		clusterUpgraderBuilder: cluster_upgrader_builder.NewBuilder(),
+		clusterUpgraderBuilder: cub.NewBuilder(),
 		validationBuilder:      validation.NewBuilder(),
 	}
 }
@@ -69,7 +72,7 @@ type ReconcileUpgradeConfig struct {
 	client                 client.Client
 	scheme                 *runtime.Scheme
 	metricsClientBuilder   metrics.MetricsBuilder
-	clusterUpgraderBuilder cluster_upgrader_builder.ClusterUpgraderBuilder
+	clusterUpgraderBuilder cub.ClusterUpgraderBuilder
 	validationBuilder      validation.ValidationBuilder
 }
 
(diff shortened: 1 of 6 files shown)


pkg/controller/upgradeconfig/upgradeconfig_controller_test.go
```diff
@@ -160,16 +160,15 @@ var _ = Describe("UpgradeConfigController", func() {
 					It("Adds it successfully", func() {
 						matcher := testStructs.NewUpgradeConfigMatcher()
 						util.ExpectGetClusterVersion(mockKubeClient, clusterVersionList, nil)
-						gomock.InOrder(
```
sedroche, line 163: `InOrder` didn't play nice with multiple `mockKubeClient.EXPECT().Status()` calls.
Happy to correct this if anyone knows how to do so correctly

pkg/controller/upgradeconfig/upgradeconfig_controller_test.go
```diff
@@ -330,16 +349,18 @@ var _ = Describe("UpgradeConfigController", func() {
...
+							mockValidationBuilder.EXPECT().NewClient().Return(mockValidator, nil)
+							mockValidator.EXPECT().IsValidUpgradeConfig(gomock.Any(), gomock.Any(), gomock.Any()).Return(true, nil)
+							mockMetricsClient.EXPECT().UpdateMetricValidationSucceeded(gomock.Any())
+							mockClusterUpgraderBuilder.EXPECT().NewClient(gomock.Any(), gomock.Any(), upgradeConfig.Spec.Type).Return(mockClusterUpgrader, nil)
+							mockClusterUpgrader.EXPECT().UpgradeCluster(gomock.Any(), gomock.Any()).Times(1).Return(upgradev1alpha1.UpgradePhaseUpgrading, &upgradev1alpha1.UpgradeCondition{}, fakeError)
+							mockKubeClient.EXPECT().Status().AnyTimes().Return(mockUpdater)
+							mockUpdater.EXPECT().Update(gomock.Any(), gomock.Any()).AnyTimes()
 							result, err := reconciler.Reconcile(reconcile.Request{NamespacedName: upgradeConfigName})
-							Expect(err).NotTo(HaveOccurred())
+							Expect(err).To(HaveOccurred())
```
sedroche, line 363: These are from:
https://github.com/openshift/managed-upgrade-operator/blob/4fd98b3f57d4eaadddc0158e336147509ccf86b8/pkg/controller/upgradeconfig/upgradeconfig_controller.go#L169-L172
we were just logging the error

pkg/controller/upgradeconfig/upgradeconfig_controller_test.go
dofinn: @sedroche can you please review this as removing `InOrder` weakens time based assertions.

pkg/osd_cluster_upgrader/upgrader_test.go
```diff
@@ -438,36 +424,16 @@ var _ = Describe("ClusterUpgrader", func() {
...
-				stepHistoryReason := upgradeConfig.Status.History.GetHistory(upgradeConfig.Spec.Desired.Version).Conditions.GetCondition(step1).Reason
-				stepHistoryMsg := upgradeConfig.Status.History.GetHistory(upgradeConfig.Spec.Desired.Version).Conditions.GetCondition(step1).Message
+			It("Indicates the error in the condition", func() {
+				_, condition, err := cu.UpgradeCluster(upgradeConfig, logger)
+				stepHistoryReason := condition.Reason
+				stepHistoryMsg := condition.Message
 				Expect(stepHistoryReason).To(Equal(string(step1) + " not done"))
 				Expect(stepHistoryMsg).To(Equal("step " + string(step1) + " failed"))
 				Expect(stepCounter[step1]).To(Equal(1))
-				// TODO - this needs eyes in cluster-upgrader - need to doublecheck whether it is expected to return nil here
```
sedroche, line 448: cc @mrbarge This was incorrect. I added a multi error here to report UpgardeCluster error and status update error:
https://github.com/sedroche/managed-upgrade-operator/blob/c8932877cc7335bf5af0dbd9f5bd1d5efff90c12/pkg/controller/upgradeconfig/upgradeconfig_controller.go#L208-L223

pkg/controller/upgradeconfig/upgradeconfig_controller.go
```diff
@@ -4,7 +4,10 @@ import (
 	"context"
 	"time"
 
+	"github.com/go-logr/logr"
+	"github.com/hashicorp/go-multierror"
 	"k8s.io/apimachinery/pkg/api/errors"
```
wanghaoran1988, line 9: move down those two lines?

pkg/controller/upgradeconfig/upgradeconfig_controller.go
```diff
@@ -180,29 +188,42 @@ func (r *ReconcileUpgradeConfig) Reconcile(request reconcile.Request) (reconcile
...
-		reqLogger.Info("Cluster detected as already upgrading.")
-		err = upgrader.UpgradeCluster(instance, reqLogger)
-		if err != nil {
-			reqLogger.Error(err, "Failed to upgrade cluster")
-		}
+		return r.upgradeCluster(upgrader, instance, reqLogger)
 	case upgradev1alpha1.UpgradePhaseUpgraded:
 		reqLogger.Info("Cluster is already upgraded")
 		return reconcile.Result{}, nil
-	case upgradev1alpha1.UpgradePhaseFailed:
```
wanghaoran1988, line 195: This is not needed anymore?

sedroche:
> Can we instead of only save the current condition, we save only maximum number of history, like only keep latest 3 upgrade history?

What is the advantage of doing so? Conditions should be "the latest available observations of an object’s state"

pkg/controller/upgradeconfig/upgradeconfig_controller.go
sedroche: I don't see it used anywhere so i removed it. We return errors so they should be logged anyway

PR #74, opened 2020-07-27
Local dev catalogsource

sedroche:
Part of https://issues.redhat.com/browse/OSD-4531 

- make command and docs to create CatalogSource

### DIFF
--- a/docs/catalogsource.md
+++ b/docs/catalogsource.md
@@ -1,0 +1,43 @@
+# Create a CatalogSource to test manifest changes
+
+## Create manifests
+```bash
+make manifests
+```
+
+## Create registry
+
+### Create registry Dockerfile
+```bash
+cat <<EOF > LocalRegistryDockerfile
+FROM quay.io/openshift/origin-operator-registry:latest
+
+COPY manifests manifests
+RUN initializer --permissive
+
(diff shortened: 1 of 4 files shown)


docs/catalogsource.md
```diff
@@ -0,0 +1,43 @@
...
+RUN initializer --permissive
+
+CMD ["registry-server", "-t", "/tmp/terminate.log"]
+EOF
+```
+
+### Build and push Dockerfile
+
+```bash
+docker build -f ./LocalRegistryDockerfile --tag quay.io/<your_repo>/managed-upgrade-operator-registry:<tag> .
```
bmeng, line 25: should we suggest to use `podman` or `imagebuilder` instead of docker?

sedroche: I'm just replicating what the app-sre scripts do. I think sticking to that is of more benefit since we will encounter any issue the same as they will do

wanghaoran1988:
Do we need remove the files under ./deploy dir ?

sedroche:
> Do we need remove the files under ./deploy dir ?

`./hack/generate-operator-bundle.py` creates the manifest based on deploy dir. I have .gitignored these files as its just for local testing

PR #75, opened 2020-07-27
Adding log messages for isReadyToUpgrade validation

ravitri:
The logging message "[Checking if cluster can commence upgrade](https://github.com/openshift/managed-upgrade-operator/blob/master/pkg/controller/upgradeconfig/upgradeconfig_controller.go#L163)" kept on repeating without mentioning what it is checking and for how long. Added relevant logging messages to address it.

NOTE: 
1. If upgrade window is breached, the controller would atleast log error messages for now but I think a validating webhook is better in this case.
2. For now, the remaining time for upgrade is displayed in hours.

### DIFF
--- a/pkg/controller/upgradeconfig/upgradeconfig_controller.go
+++ b/pkg/controller/upgradeconfig/upgradeconfig_controller.go
@@ -236,7 +236,7 @@ func isReadyToUpgrade(upgradeConfig *upgradev1alpha1.UpgradeConfig, metricsClien
 	}
 	upgradeTime, err := time.Parse(time.RFC3339, upgradeConfig.Spec.UpgradeAt)
 	if err != nil {
-		log.Error(err, "Failed to parse spec.upgradeAt", upgradeConfig.Spec.UpgradeAt)
+		log.Error(err, "failed to parse spec.upgradeAt", upgradeConfig.Spec.UpgradeAt)
 		return false
 	}
 	now := time.Now()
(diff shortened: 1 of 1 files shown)


pkg/controller/upgradeconfig/upgradeconfig_controller.go
```diff
@@ -226,25 +231,29 @@ func (r *ReconcileUpgradeConfig) upgradeCluster(upgrader cub.ClusterUpgrader, uc
...
 func isReadyToUpgrade(upgradeConfig *upgradev1alpha1.UpgradeConfig, metricsClient metrics.Metrics) bool {
 	if !upgradeConfig.Spec.Proceed {
-		log.Info("upgrade cannot be proceed", "proceed", upgradeConfig.Spec.Proceed)
+		log.Info("Upgrade cannot proceed", "proceed", upgradeConfig.Spec.Proceed)
 		return false
 	}
 	upgradeTime, err := time.Parse(time.RFC3339, upgradeConfig.Spec.UpgradeAt)
 	if err != nil {
-		log.Error(err, "failed to parse spec.upgradeAt", upgradeConfig.Spec.UpgradeAt)
+		log.Error(err, "Failed to parse spec.upgradeAt", upgradeConfig.Spec.UpgradeAt)
```
wanghaoran1988, line 239: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

ravitri: Thanks Haoran! Good to know about difference in standards for Error and Info msgs.. Reverted it..

PR #77, opened 2020-07-27
trigger the alert if node cannot be drained in the specified duration

bmeng:
https://issues.redhat.com/browse/OSD-4528

### DIFF
--- a/pkg/osd_cluster_upgrader/upgrader.go
+++ b/pkg/osd_cluster_upgrader/upgrader.go
@@ -256,11 +256,9 @@ func AllWorkersUpgraded(c client.Client, scaler scaler.Scaler, metricsClient met
 		return false, errSilence
 	}
 
-	if !okUpgrade {
-		if !silenceActive {
-			logger.Info("Worker upgrade timeout.")
-			metricsClient.UpdateMetricUpgradeWorkerTimeout(upgradeConfig.Name, upgradeConfig.Spec.Desired.Version)
-		}
+	if !okUpgrade && !silenceActive {
+		logger.Info("Worker upgrade timeout.")
+		metricsClient.UpdateMetricUpgradeWorkerTimeout(upgradeConfig.Name, upgradeConfig.Spec.Desired.Version)
 		return false, nil
 	}
 
--- a/pkg/osd_cluster_upgrader/upgrader.go
+++ b/pkg/osd_cluster_upgrader/upgrader.go
@@ -235,17 +235,28 @@ func CreateWorkerMaintWindow(c client.Client, scaler scaler.Scaler, metricsClien
 
 // This check whether all the worker nodes are ready with new config
 func AllWorkersUpgraded(c client.Client, scaler scaler.Scaler, metricsClient metrics.Metrics, m maintenance.Maintenance, upgradeConfig *upgradev1alpha1.UpgradeConfig, logger logr.Logger) (bool, error) {
-	ok, err := nodesUpgraded(c, "worker", logger)
-	if err != nil {
-		return false, err
+	okDrain, errDrain := nodeDrained(c, upgradeConfig, logger)
+	if errDrain != nil {
+		return false, errDrain
 	}
 
-	silenceActive, err := m.IsActive()
-	if err != nil {
-		return false, err
+	if !okDrain {
+		logger.Info("Node drain timeout.")
+		metricsClient.UpdateMetricNodeDrainFailed(upgradeConfig.Name)
+		return false, nil
 	}
 
-	if !ok {
+	okUpgrade, errUpgrade := nodesUpgraded(c, "worker", logger)
+	if errUpgrade != nil {
+		return false, errUpgrade
+	}
+
+	silenceActive, errSilence := m.IsActive()
+	if errSilence != nil {
+		return false, errSilence
+	}
+
+	if !okUpgrade {
 		if !silenceActive {
 			logger.Info("Worker upgrade timeout.")
 			metricsClient.UpdateMetricUpgradeWorkerTimeout(upgradeConfig.Name, upgradeConfig.Spec.Desired.Version)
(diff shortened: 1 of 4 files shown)


pkg/osd_cluster_upgrader/upgrader.go
```diff
@@ -567,3 +578,28 @@ func hasUpgradeCommenced(c client.Client, uc *upgradev1alpha1.UpgradeConfig) (bo
 
 	return true, nil
 }
+
+func nodeDrained(c client.Client, uc *upgradev1alpha1.UpgradeConfig, logger logr.Logger) (bool, error) {
```
wanghaoran1988, line 582: We are not checking node/nodes drained here, because the node upgrade is one by one, better rename this to `hasStuckNode` maybe?

pkg/osd_cluster_upgrader/upgrader.go
```diff
@@ -567,3 +578,28 @@ func hasUpgradeCommenced(c client.Client, uc *upgradev1alpha1.UpgradeConfig) (bo
...
+
+	drainTimeout := time.Duration(uc.Spec.PDBForceDrainTimeout) * time.Minute
+	var drainStarted metav1.Time
+	for _, node := range nodeList.Items {
+		if node.Spec.Unschedulable && len(node.Spec.Taints) > 0 {
+			for _, n := range node.Spec.Taints {
+				if n.Effect == "NoSchedule" {
+					drainStarted = *n.TimeAdded
+					if drainStarted.Add(drainTimeout).Before(metav1.Now().Time) {
+						logger.Info("Drain node failed")
```
wanghaoran1988, line 597: logger.Info("The node cannot be drained within * minutes") ?

pkg/osd_cluster_upgrader/upgrader.go
wanghaoran1988: Also please add a comment for the func

bmeng: the node is being drained during the upgrade process by the machine-config-operator 
https://github.com/openshift/machine-config-operator/blob/master/pkg/daemon/update.go#L144

the node is marked as unschedulable when the drain started, and mark the start time in the spec.taints. And marked it as schedulable again after the drain finished.
The function here is checking that the node is marking as unschedulable longer than given duration.

pkg/osd_cluster_upgrader/upgrader.go
```diff
@@ -567,3 +580,29 @@ func hasUpgradeCommenced(c client.Client, uc *upgradev1alpha1.UpgradeConfig) (bo
...
+	if errNode != nil {
+		return false, errNode
+	}
+
+	drainTimeout := time.Duration(uc.Spec.PDBForceDrainTimeout) * time.Minute
+	var drainStarted metav1.Time
+	for _, node := range nodeList.Items {
+		if node.Spec.Unschedulable && len(node.Spec.Taints) > 0 {
+			for _, n := range node.Spec.Taints {
+				if n.Effect == "NoSchedule" {
```
sedroche, line 597:
```suggestion
				if n.Effect == corev1.TaintEffectNoSchedule {
```

pkg/osd_cluster_upgrader/upgrader.go
```diff
@@ -567,3 +580,35 @@ func hasUpgradeCommenced(c client.Client, uc *upgradev1alpha1.UpgradeConfig) (bo
...
+
+// nodeDrained checks if the nodes are being drained successfully during the upgrade
+func nodeDrained(c client.Client, uc *upgradev1alpha1.UpgradeConfig, logger logr.Logger, metrics metrics.Metrics) (bool, error) {
+	nodeList := &corev1.NodeList{}
+	errNode := c.List(context.TODO(), nodeList)
+	if errNode != nil {
+		return false, errNode
+	}
+
+	pdbAlertName := "PodDisruptioinBudgetAtTime"
```
sedroche, line 592: `PodDisruptionBudgetAtTime`

ravitri: @bmeng - Did you intend to refer to [PodDisruptionBudgetAtLimit](https://github.com/openshift/cluster-kube-controller-manager-operator/blob/master/manifests/0000_90_kube-controller-manager-operator_05_alert-pdb.yaml#L12) alert here instead of `PodDisruptionBudgetAtTime` ?

bmeng: thanks for catching the stupid typo...

mrbarge:
@bmeng can you please create a follow-up card to cover creating test case(s) for this functionality? (if not going to be a part of this PR)

bmeng:
@mrbarge https://issues.redhat.com/browse/OSD-4639 The card for testing created.

pkg/osd_cluster_upgrader/upgrader.go
```diff
@@ -567,3 +580,35 @@ func hasUpgradeCommenced(c client.Client, uc *upgradev1alpha1.UpgradeConfig) (bo
...
+// nodeDrained checks if the nodes are being drained successfully during the upgrade
+func nodeDrained(c client.Client, uc *upgradev1alpha1.UpgradeConfig, logger logr.Logger, metrics metrics.Metrics) (bool, error) {
+	nodeList := &corev1.NodeList{}
+	errNode := c.List(context.TODO(), nodeList)
+	if errNode != nil {
+		return false, errNode
+	}
+
+	pdbAlertName := "PodDisruptionBudgetAtLimit"
+	pdbAlerts, errMetric := metrics.Query(fmt.Sprintf("ALERTS{alertname=\"%s\"}", pdbAlertName))
```
dofinn, line 593: So if we only allow Critical alerts to be raised during an upgrade. This will never be detected as it `severity:warning` as per: https://github.com/openshift/cluster-kube-controller-manager-operator/blob/master/manifests/0000_90_kube-controller-manager-operator_05_alert-pdb.yaml#L12

We ignore `warnings` in our own silences: https://github.com/openshift/managed-upgrade-operator/blob/master/pkg/maintenance/alertmanagerMaintenance.go#L185

bmeng: The silence was created in Alertmanager, where this call is calling the prometheus.

The prometheus will always fire the alert with specified severity, the alertmanager will decide if alert with integrated tools like PagerDuty.

So it should not be a problem for us.

sedroche: I was thinking about this too and I think its better to check for `PodDisruptionBudget` objects directly instead of querying prometheus. This removes prometheus and keeps it Kubernetes.
Looking at the alert it checks `status_expected_pods == status_desired_healthy`

bmeng: @sedroche Moved the pdb detection out of prometheus

PR #78, opened 2020-07-27
Improve PR check

ravitri:
Add checks to app-sre PR script to run `make generate` and `make lint` for every PR that is raised. It should fail if any local changes found after `make generate` command. `golangci-lint cache clean` is added for local run of `make lint` command failures.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,4 +36,5 @@ run:
 
 .PHONY: lint
 lint:
+	golangci-lint cache clean
 	golangci-lint run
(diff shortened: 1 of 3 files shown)


2uasimojo:
It looks like the regenerated file results in a circular import. That's certainly damage from a different PR, and should arguably be fixed in a different PR, but this will need to be respun either way.

Leaving my approval, since this is doing the right thing, and won't merge until CI passes anyway. (Not to mention that I'm not an approver in this repo, so my lgtm doesn't mean much :P )

sedroche:
Can we add a gofmt check too? `gofmt -l ./pkg/ util/` maybe?

ravitri:
/close
To be handled as part of the boilerplate project's standardization

PR #80, opened 2020-07-28
Adding make generate changes for validation package

ravitri:
Reverted this commit as part of https://github.com/openshift/managed-upgrade-operator/pull/78 and raising separate PR since #78 would require separate effort. 

The change in this PR is result of running `make generate` locally after a rebase since it was missed as part of some other PR. #78 is intended to address this gap for long term.

### DIFF
--- a/pkg/validation/mockValidationBuilder.go
+++ b/pkg/validation/mockValidationBuilder.go
@@ -6,6 +6,7 @@ package validation
 
 import (
 	gomock "github.com/golang/mock/gomock"
+	validation "github.com/openshift/managed-upgrade-operator/pkg/validation"
 	reflect "reflect"
 )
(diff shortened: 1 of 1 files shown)


pkg/validation/mockValidationBuilder.go
```diff
@@ -6,6 +6,7 @@ package validation
 
 import (
 	gomock "github.com/golang/mock/gomock"
+	validation "github.com/openshift/managed-upgrade-operator/pkg/validation"
```
sedroche, line 9: Can you double check this?
Shouldn't need to import the same pkg

ravitri:
Closing PR as discussed with John.. The fix will be part of another PR

PR #95, opened 2020-08-03
Restore InOrder to tests

sedroche:
Jira: https://issues.redhat.com/browse/OSD-4653

### DIFF
--- a/pkg/controller/upgradeconfig/upgradeconfig_controller.go
+++ b/pkg/controller/upgradeconfig/upgradeconfig_controller.go
@@ -173,8 +173,8 @@ func (r *ReconcileUpgradeConfig) Reconcile(request reconcile.Request) (reconcile
 		}
 
 		reqLogger.Info("Checking if cluster can commence upgrade.")
-		ready := r.scheduler.IsReadyToUpgrade(instance, metricsClient, cfg.GetUpgradeWindowTimeOutDuration())
-		if ready {
+		result := r.scheduler.IsReadyToUpgrade(instance, cfg.GetUpgradeWindowTimeOutDuration())
+		if result.IsReady {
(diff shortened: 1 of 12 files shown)


sedroche:
> This is a big improvement. However, it doesn't look like you're finalizing your mock controller, which you still need to do explicitly in go 1.13. AFAIU, when using Ginkgo, you want to do that in a `[Just]AfterEach` corresponding to the `[Just]BeforeEach` where you're initializing the controller (in your case L61 in `BeforeEach`). Here's an [example](https://github.com/openshift/aws-account-operator/blob/93fdb6ab6369678c90685a6571022de7da090f4e/pkg/controller/accountclaim/accountclaim_controller_test.go#L70-L72).
> 
> Doing so should allow you to clean up a bunch more things. For example:
> 
> * `.Times(1)` is redundant, it's the default.
> * `.Times(0)` should be the same as omitting the check entirely -- though including it as an explicit statement of "this thing doesn't happen" is nice for readability.
> * I would discourage the use of `AnyTimes()` if possible, unless you're truly expecting that call to happen an _unpredictable_ number of times in a row in that spot in the sequence.

Thanks @2uasimojo , applying this feedback showed up some incorrect tests which have been fixed now as well.
We went with removing `.Times(1)` and `AnyTimes()` but keeping `Times(0)` for readability.

sedroche:
CC @dofinn & @mrbarge This ended up being quite a big change but its mostly the test feedback from Eric.
Just calling out https://github.com/openshift/managed-upgrade-operator/pull/95/commits/16b804f8d6b6f180af62092ec8fda78b05623817 as a change to the controller code

2uasimojo:
> Thanks @2uasimojo , applying this feedback showed up some incorrect tests which have been fixed now as well.

Awesome! Sorry I didn't get back to reviewing before it merged, but glad to hear this.

PR #96, opened 2020-08-05
Refine validation error handling

sedroche:
Validation was setting the failed validation metric on err.

- I have refined the flow to only set validation metric on validation failure
- One new test for setting the failed alert metric correctly

### DIFF
--- a/pkg/controller/upgradeconfig/upgradeconfig_controller_test.go
+++ b/pkg/controller/upgradeconfig/upgradeconfig_controller_test.go
@@ -239,7 +239,7 @@ var _ = Describe("UpgradeConfigController", func() {
 					}
 				})
 				Context("When the upgradeconfig validation fails", func() {
-					It("should set set the validation alert metric", func() {
+					It("should set the validation alert metric", func() {
 						util.ExpectGetClusterVersion(mockKubeClient, clusterVersionList, nil)
 
 						gomock.InOrder(
--- a/pkg/controller/upgradeconfig/upgradeconfig_controller_test.go
+++ b/pkg/controller/upgradeconfig/upgradeconfig_controller_test.go
@@ -22,6 +22,7 @@ import (
 	mockMetrics "github.com/openshift/managed-upgrade-operator/pkg/metrics/mocks"
 	"github.com/openshift/managed-upgrade-operator/pkg/scheduler"
 	schedulerMocks "github.com/openshift/managed-upgrade-operator/pkg/scheduler/mocks"
+	"github.com/openshift/managed-upgrade-operator/pkg/validation"
 	validationMocks "github.com/openshift/managed-upgrade-operator/pkg/validation/mocks"
 	"github.com/openshift/managed-upgrade-operator/util"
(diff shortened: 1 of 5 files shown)


sedroche:
cc @dofinn

pkg/validation/validation.go
```diff
@@ -25,56 +25,94 @@ func NewBuilder() ValidationBuilder {
...
 type Validator interface {
-	IsValidUpgradeConfig(uC *upgradev1alpha1.UpgradeConfig, cV *configv1.ClusterVersion, logger logr.Logger) (bool, error)
+	IsValidUpgradeConfig(uC *upgradev1alpha1.UpgradeConfig, cV *configv1.ClusterVersion, logger logr.Logger) (ValidatorResult, error)
 }
 
 type validator struct{}
 
-func (v *validator) IsValidUpgradeConfig(uC *upgradev1alpha1.UpgradeConfig, cV *configv1.ClusterVersion, logger logr.Logger) (bool, error) {
+type ValidatorResult struct {
+	IsRollBack bool
```
dofinn, line 34: Where is this used? What is its purpose?

wanghaoran1988:
If first time the validation failed for some reason, and then succeed next time, will the metrics be updated? if not, do we have alert use this metrics? and will the alert be cleared?

sedroche:
> If first time the validation failed for some reason, and then succeed next time, will the metrics be updated? if not, do we have alert use this metrics? and will the alert be cleared?

I believe the prometheus rules only fire after 5 mins:
https://github.com/openshift/managed-cluster-config/blob/88e04b670c3b3bd05416143ef2e1271720bab9a2/deploy/sre-prometheus/100-managed-upgrade-operator.PrometheusRule.yaml#L16

we reset the metric after its validated:
https://github.com/openshift/managed-upgrade-operator/blob/86c096884438cec038b431017bec372ce614336b/pkg/controller/upgradeconfig/upgradeconfig_controller.go#L166

pkg/validation/validation.go
sedroche: Good spot.
There was a scenario of `false, nil` for roll back which I tried to capture.
Ok to remove iyo?

pkg/controller/upgradeconfig/upgradeconfig_controller_test.go
```diff
@@ -237,6 +238,20 @@ var _ = Describe("UpgradeConfigController", func() {
 						},
 					}
 				})
+				Context("When the upgradeconfig validation fails", func() {
+					It("should set set the validation alert metric", func() {
```
bmeng, line 242: dupe "set"

pkg/validation/validation.go
bmeng: Will we support rollback? Do we need to separate this condition out?

sedroche: @bmeng @dofinn I'm going to remove this

pkg/controller/upgradeconfig/upgradeconfig_controller_test.go
sedroche: Thank you

PR #118, opened 2020-09-04
ClusterVersion client

sedroche:
### What type of PR is this?
Refactor

- Adding new ClusterVersion pkg to centralise interactions with the CV/CVO which may be used across different upgraders/controllers 
- Associated Test updates

This PR is just the initial pkg and moving one function to the pkg. There are others functions which will be moved but to keep the PRs manageable this is the first piece of standalone work.
Also I have attempted to split the commits into logical commits to review easier.


pkg/clusterversion/cv.go
```diff
@@ -0,0 +1,58 @@
...
+	"context"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	OSD_CV_NAME string = "version"
```
bmeng, line 13: Should use camel format here?
And looks like this variable is not being used in here. Should it be declared somewhere else?
Another question is this should be a `const` or `var`?

sedroche: Yeah, im in 2 minds about the way we do this. I'll remove it for now. Thanks @bmeng

mrbarge:
Build and tested on a cluster, ran local tests, all fine.
/lgtm

PR #122, opened 2020-09-07
Removing imports left over from previous fake client

sedroche:
### What type of PR is this?
cleanup

These were left over from when we used the k8s client fake pkg before moving to ginko

- [x] Ran make test


mrbarge:
Built and tested locally, all good.
/lgtm

PR #125, opened 2020-09-07
lowercase pdb field

dofinn:
### What type of PR is this?
cleanup

Standardisation with rest of repo/project.

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ metadata:
 spec:
   type: "OSD"
   upgradeAt: "2020-01-01T00:00:00Z"
-  PDBForceDrainTimeout: 60
+  pdbForceDrainTimeout: 60
   desired:
     channel: "fast-4.4"
     version: "4.4.6"
--- a/deploy/crds/upgrade.managed.openshift.io_upgradeconfigs_crd.yaml
+++ b/deploy/crds/upgrade.managed.openshift.io_upgradeconfigs_crd.yaml
@@ -53,12 +53,6 @@ spec:
           description: UpgradeConfigSpec defines the desired state of UpgradeConfig
             and upgrade window and freeze window
           properties:
-            PDBForceDrainTimeout:
-              description: The maximum grace period granted to a node whose drain
-                is blocked by a Pod Disruption Budget, before that drain is forced.
-                Measured in minutes.
-              format: int32
-              type: integer
             desired:
               description: Specify the desired OpenShift release
               properties:
@@ -72,6 +66,12 @@ spec:
               - channel
               - version
               type: object
+            pdbForceDrainTimeout:
+              description: The maximum grace period granted to a node whose drain
+                is blocked by a Pod Disruption Budget, before that drain is forced.
+                Measured in minutes.
+              format: int32
+              type: integer
             subscriptionUpdates:
               description: This defines the 3rd party operator subscriptions upgrade
               items:
@@ -103,8 +103,8 @@ spec:
               description: Specify the upgrade start time
               type: string
           required:
-          - PDBForceDrainTimeout
           - desired
+          - pdbForceDrainTimeout
           - type
           - upgradeAt
           type: object
--- a/deploy/crds/upgrade.managed.openshift.io_v1alpha1_upgradeconfig_cr.yaml
+++ b/deploy/crds/upgrade.managed.openshift.io_v1alpha1_upgradeconfig_cr.yaml
@@ -5,7 +5,7 @@ metadata:
 spec:
   type: "OSD"
   upgradeAt: "2020-01-01T00:00:00Z"
-  PDBForceDrainTimeout: 60
+  pdbForceDrainTimeout: 60
   desired:
     channel: "fast-4.4"
     version: "4.3.19"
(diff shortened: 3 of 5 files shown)


mrbarge:
We need to manage this carefully - we have clusters with UpgradeConfigs already on them that feature a capitalized PDB. I believe this change will result in those clusters now reading in a PDB Force Drain Timeout of 0 because it is expecting a lowercase "pdb" field in the JSON and won't find it, so it will leave the default of 0 when unmarshalling. I was able to verify this with a quick test.

sedroche:
> We need to manage this carefully - we have clusters with UpgradeConfigs already on them that feature a capitalized PDB. I believe this change will result in those clusters now reading in a PDB Force Drain Timeout of 0 because it is expecting a lowercase "pdb" field in the JSON and won't find it, so it will leave the default of 0 when unmarshalling. I was able to verify this with a quick test.

Did you test with a replacement manifest?

mrbarge:
> > We need to manage this carefully - we have clusters with UpgradeConfigs already on them that feature a capitalized PDB. I believe this change will result in those clusters now reading in a PDB Force Drain Timeout of 0 because it is expecting a lowercase "pdb" field in the JSON and won't find it, so it will leave the default of 0 when unmarshalling. I was able to verify this with a quick test.
> 
> Did you test with a replacement manifest?

No, I just ran the operator locally, so still using the existing one. Either way, I would like some testing first on existing clusters with previous versions of the operator/UC to ensure this isn't going to some trouble when promoted.

sedroche:
> Either way, I would like some testing first on existing clusters that have been running previous versions of the operator/UC to ensure this isn't going to some trouble when promoted.

Agreed. fyi there is [a doc](https://github.com/openshift/managed-upgrade-operator/blob/master/docs/catalogsource.md) for creating manifests to test.

mrbarge:
Gave this a test and it seemed to behave okay. My testing scenario was:

- Created cluster with current prod version of MUO.
- Synced an UpgradeConfig (with uppercase PDB) to the cluster via Hive. 
- Upgraded cluster.
- Deployed this branch of MUO. Update the CRD to the one in this branch.

At this point, I observed no validation failures or alerts, however, MUO reported a "0" for the PDB config value as anticipated.
I also observed that, at this point, the existing Hive syncset of the UpgradeConfig no longer continued to sync, because the CR was invalid (an uppercase PDB).

- I updated the Hive syncset to use a lowercase PDB. Syncing of the Syncset resumed.
- MUO reported "60" for the PDB config value.

Based on these tests I feel pretty confident a promotion of this won't hurt production, though, we do need to do a regeneration of our SyncSets shortly after promotion.

@dofinn @sedroche if you are satisfied with the findings above I will remove my hold and approve.

sedroche:
I would like to test this by manifest replacement using OLM to be 100% sure but the rest of this seems fine

mrbarge:
So I did a test via manifest replacement as per @sedroche suggestion.

- Created a cluster with a custom catalogsource running the master branch MUO.
- Created an upgradeconfig with `PDBForceDrainTimeout` (uppercase)
- Updated the CatalogSource to install a new version of MUO featuring this branch.

When I approved the InstallPlan, the installation **failed** because the existing `UpgradeConfig` did not have a lowercase `pdbForceDrainTimeout` field:

```
  - lastTransitionTime: "2020-09-21T01:14:22Z"                                                                                                                                                                         lastUpdateTime: "2020-09-21T01:14:22Z"                                                                                                                                                                         
    message: 'error validating existing CRs agains new CRD''s schema: upgradeconfigs.upgrade.managed.openshift.io:                                                                                                 
      error validating custom resource against new schema &apiextensions.CustomResourceValidation{OpenAPIV3Schema:(*apiextensions.JSONSchemaProps)(0xc0022301e0)}:                                                 
      [].spec.pdbForceDrainTimeout: Required value'                                                                                                                                                                
```

I then tried a second test, this time ensuring that the UpgradeConfig had both an upper and lowercase variant of the `PDBForceDrainTimeout` field prior to approving the InstallPlan to upgrade MUO. This time the upgrade completed successfully.

Therefore if we go ahead with this change we **must** ensure all clusters have received a new UpgradeConfig that contains an uppercase and lowercase version of the `PDBForceDrainTimeout` field first, or else the installation will fail.

sedroche:
> Therefore if we go ahead with this change we **must** ensure all clusters have received a new UpgradeConfig that contains an uppercase and lowercase version of the `PDBForceDrainTimeout` field first, or else the installation will fail.

I presume it can just be lowercase? Can we sync this to when/if this lands? -> https://github.com/openshift/managed-upgrade-operator/pull/132#issuecomment-694529366

mrbarge:
> > Therefore if we go ahead with this change we **must** ensure all clusters have received a new UpgradeConfig that contains an uppercase and lowercase version of the `PDBForceDrainTimeout` field first, or else the installation will fail.
> 
> I presume it can just be lowercase? Can we sync this to when/if this lands? -> [#132 (comment)](https://github.com/openshift/managed-upgrade-operator/pull/132#issuecomment-694529366)

You're right, if it just has a lowercase version, then the old version of MUO will still be fine, it'll just read a value of 0 until a version of MUO with this branch lands.

I'll drop my thinking about how we manage this from an operational perspective on Slack.

mrbarge:
@dofinn can you please fix the merge conflict here - I will then give it another test with a custom catalogsource and make sure this is good to go live.

PR #128, opened 2020-09-11
Alert on node drain controller

sedroche:
### What type of PR is this?
Refactor

To really watch on nodes draining this needs to happen per node and in parallel to master machineconfigpools as workers update at the same time as masters
This PR uses the sdk controller pattern to watch for node events and will alert if a node drain times out.
This is the first step in implementing enhanced node draining and should be feature parity with the previous implementation.
Using the controller pattern allows us to control draining outside of upgraders.

https://issues.redhat.com/browse/OSD-4716
This does not implement the force drain as mention in the above jira, that will be a follow on PR

Basic ask is to alert SRE if a operator initiated worker node passes our node drain timeout configured in managed-upgrade-operator-config configmap

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -56,7 +56,7 @@ $ oc new-project managed-upgrade-operator
 * Run the operator via the Operator SDK:
 
 ```
-$ operator-sdk run --local --watch-namespace=""
+$ OPERATOR_NAMESPACE=managed-upgrade-operator operator-sdk run --local --watch-namespace=""
 ``` 
 
 (`make run` will also work here)
--- a/pkg/controller/nodekeeper/nodekeeper_controller.go
+++ b/pkg/controller/nodekeeper/nodekeeper_controller.go
@@ -1,0 +1,195 @@
+package nodekeeper
+
+import (
+	"context"
+	"fmt"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"os"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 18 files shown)


docs/development.md
```diff
@@ -56,7 +56,7 @@ $ oc new-project managed-upgrade-operator
 * Run the operator via the Operator SDK:
 
 ```
-$ operator-sdk run --local --watch-namespace=""
+$ OPERATOR_NAMESPACE=managed-upgrade-operator operator-sdk run --local --watch-namespace=""
```
ravitri, line 59: The namespace is openshift-managed-upgrade-operator right?

sedroche: Seems to just follow what is created in the create project step -> `oc new-project managed-upgrade-operator`

ravitri: I think it would still work and be fine to mention some other namespace for the upgradeconfig resource to be present but perhaps even the repo docs can uniformly maintain the same namespace name as in the deployment of the operator https://github.com/openshift/managed-upgrade-operator/blob/nodekeeper_controller/deploy/namespace.yaml#L4

pkg/controller/nodekeeper/nodekeeper_controller.go
```diff
@@ -0,0 +1,195 @@
...
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, err
+	}
+
+	result := isDraining(node)
+	metricsClient, err := r.metricsClientBuilder.NewClient(r.client)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	if !result.IsDraining {
```
dofinn, line 125: did we need to evaluate this the same as https://github.com/openshift/managed-upgrade-operator/pull/128/files#diff-52b33f64f27b313ab7028c0fa1db2024R105

dofinn (commented):
LGTM, will let @mrbarge sign off officially.

docs/development.md
sedroche: Makes sense. I'll update that :+1:

mrbarge:
I do intend to invest more time in reviewing this more thoroughly, but my initial 1-pass over it was fine and I'd rather not hold it up.
/lgtm

PR #136, opened 2020-09-22
Lockdown watch namespace

sedroche:
Ensures the operator only watches on the ns its deployed too


mrbarge:
Tested and seems happy.
/lgtm

PR #137, opened 2020-09-23
fix the worker maint logic to stop keep creating maint

bmeng:
### What type of PR is this?
bug

fixes OSD-5090

### DIFF
--- a/pkg/maintenance/alertmanagerMaintenance.go
+++ b/pkg/maintenance/alertmanagerMaintenance.go
@@ -138,31 +138,36 @@ func (amm *alertManagerMaintenance) StartControlPlane(endsAt time.Time, version 
 
 // Start a worker node maintenance in Alertmanager for version
 // Time is converted to UTC
-func (amm *alertManagerMaintenance) SetWorker(endsAt time.Time, version string) error {
+func (amm *alertManagerMaintenance) SetWorker(endsAt time.Time, version string, count int32) error {
 	comment := fmt.Sprintf("Silence for %s upgrade to version %s", workerSilenceCommentId, version)
-	silenceList, err := amm.client.Filter(equalsComment(comment))
+	fullComment := fmt.Sprintf("%s with remaining %d nodes", comment, count)
+	silenceList, err := amm.client.Filter(equalsComment(fullComment))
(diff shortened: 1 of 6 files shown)


pkg/maintenance/alertmanagerMaintenance.go
```diff
@@ -138,31 +138,36 @@ func (amm *alertManagerMaintenance) StartControlPlane(endsAt time.Time, version
 
 // Start a worker node maintenance in Alertmanager for version
 // Time is converted to UTC
-func (amm *alertManagerMaintenance) SetWorker(endsAt time.Time, version string) error {
+func (amm *alertManagerMaintenance) SetWorker(endsAt time.Time, version string, count int32) error {
```
sedroche, line 141: Does the same logic work if you compare the endAt time instead of passing the count?

pkg/maintenance/maintenance_test.go
```diff
@@ -46,7 +49,7 @@ var _ = Describe("Alert Manager Maintenance Client", func() {
 
 		activeSilenceId      = "test-id"
 		activeSilenceStatus  = amv2Models.SilenceStatusStateActive
-		activeSilenceComment = "Silence for OSD worker node upgrade to version " + testVersion
+		activeSilenceComment = "Silence for OSD with " + strconv.Itoa(int(testWorkerCount)) + " worker node upgrade to version " + testVersion
```
sedroche, line 52: we use` fmt.SprintF` elsewhere, can we use that here too?

pkg/maintenance/alertmanagerMaintenance.go
bmeng: We pass the `endAt` as a dynamic value "time.Now().Add(xxx)" in the reconcile loop. And we cannot store the value in anywhere.
I do not think it is possible to find the exact silence by the `endAt`

sedroche: Gotcha :+1:

mrbarge:
Great change. Tested it on an upgrade, verified that it only resulted in the expected number of expired worker silences afterwards.
/lgtm

PR #138, opened 2020-09-24
ignore the logging component in the pre/post health check

bmeng:
### What type of PR is this?
feature

OSD-5107

### DIFF
--- a/deploy/managed-upgrade-operator-config.yaml
+++ b/deploy/managed-upgrade-operator-config.yaml
@@ -26,3 +26,10 @@ data:
       - MetricsClientSendFailingSRE
       - UpgradeNodeScalingFailedSRE
       - UpgradeClusterCheckFailedSRE
+    verification:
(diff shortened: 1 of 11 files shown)


mrbarge:
Tested this, pretty happy that it's fine. However I think we still have an edge case to deal with in this card.

In the PostUpgradeVerification we check all daemonsets and replicasets in the openshift* namespace. If logging is broken it will fail postupgradeverification and never complete the upgrade because the daemonsets of fluentd and/or replicasets of elasticsearch may not be satisfied.

So I think we should be ignoring openshift-logging* in the PostUpgradeVerification. Just double-check with @penguinCCCP because I think I've asked him to look into turning the namespaces into a configurable item in his refactoring card, so he may already be working on this bit of code.

bmeng:
unexpected commits included, replaced by https://github.com/openshift/managed-upgrade-operator/pull/139

PR #139, opened 2020-09-25
ignore the logging component in the health check

bmeng:
### What type of PR is this?
feature

_Fixes OSD-5017

Replace https://github.com/openshift/managed-upgrade-operator/pull/138

### DIFF
--- a/pkg/osd_cluster_upgrader/upgrader.go
+++ b/pkg/osd_cluster_upgrader/upgrader.go
@@ -532,7 +532,7 @@ func performClusterHealthCheck(c client.Client, metricsClient metrics.Metrics, c
 	if len(ic) > 0 {
 		icQuery = `,alertname!="` + strings.Join(ic, `",alertname!="`) + `"`
 	}
-	healthCheckQuery := `ALERTS{alertstate="firing",severity="critical",namespace=~"^openshift.*|^kube.*|^default$",namespace!="openshift-customer-monitoring",pod!~"^elasticsearch-.*",service!="fluentd"` + icQuery + "}"
+	healthCheckQuery := `ALERTS{alertstate="firing",severity="critical",namespace=~"^openshift.*|^kube.*|^default$",namespace!="openshift-customer-monitoring|openshift-logging"` + icQuery + "}"
 	alerts, err := metricsClient.Query(healthCheckQuery)
 	if err != nil {
 		return false, fmt.Errorf("Unable to query critical alerts: %s", err)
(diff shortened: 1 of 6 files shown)


pkg/osd_cluster_upgrader/upgrader.go
```diff
@@ -395,10 +398,12 @@ func performUpgradeVerification(c client.Client, metricsClient metrics.Metrics,
 	totalDS := 0
 	for _, daemonSet := range daemonSetList.Items {
 		for _, namespacePrefix := range namespacePrefixesToCheck {
-			if strings.HasPrefix(daemonSet.Namespace, namespacePrefix) {
-				totalDS = totalDS + 1
-				if daemonSet.Status.DesiredNumberScheduled == daemonSet.Status.NumberReady {
-					readyDS = readyDS + 1
+			for _, ignoredNS := range namespaceToIgnore {
+				if strings.HasPrefix(daemonSet.Namespace, namespacePrefix) && daemonSet.Namespace != ignoredNS {
```
mrbarge, line 402: This is fine and should work, but I think it might lead to an inflated count of the number of daemonsets when reported in the subsequent log. (it'll be 'actual number of daemonsets' * 'number of ignored namespaces'). 
Just a very small thing that only really matters when it comes to the resulting log. I would be fine if it's not dealt with.

pkg/osd_cluster_upgrader/upgrader.go
```diff
@@ -525,7 +532,7 @@ func performClusterHealthCheck(c client.Client, metricsClient metrics.Metrics, c
 	if len(ic) > 0 {
 		icQuery = `,alertname!="` + strings.Join(ic, `",alertname!="`) + `"`
 	}
-	healthCheckQuery := `ALERTS{alertstate="firing",severity="critical",namespace=~"^openshift.*|^kube.*|^default$",namespace!="openshift-customer-monitoring"` + icQuery + "}"
+	healthCheckQuery := `ALERTS{alertstate="firing",severity="critical",namespace=~"^openshift.*|^kube.*|^default$",namespace!="openshift-customer-monitoring",pod!~"^elasticsearch-.*",service!="fluentd"` + icQuery + "}"
```
mrbarge, line 535: I don't think this is enough to catch all the types of logging failures. I've tested this PR with a few scenarios where logging is 'broken' and the upgrader still fails the health check:
 - Logging pods can't schedule
 - Logging pods never become ready
 - Logging volume fills to a warning/critical level
In these scenarios alerts will fire in the "openshift-logging" namespace that are not related to elasticsearch or fluentd pods/services (in the case of the former two alerts, the pod/service is kube-state-metrics). 

Maybe it just needs to ignore the openshift-logging namespace altogether? Thoughts from anyone else?

pkg/osd_cluster_upgrader/upgrader.go
```diff
@@ -525,7 +532,7 @@ func performClusterHealthCheck(c client.Client, metricsClient metrics.Metrics, c
 	if len(ic) > 0 {
 		icQuery = `,alertname!="` + strings.Join(ic, `",alertname!="`) + `"`
 	}
-	healthCheckQuery := `ALERTS{alertstate="firing",severity="critical",namespace=~"^openshift.*|^kube.*|^default$",namespace!="openshift-customer-monitoring"` + icQuery + "}"
+	healthCheckQuery := `ALERTS{alertstate="firing",severity="critical",namespace=~"^openshift.*|^kube.*|^default$",namespace!="openshift-customer-monitoring|openshift-logging"` + icQuery + "}"
```
mrbarge, line 535: Still having problems with this query. It needs to use the regex exquivalent of != as in the suggestion below:
```suggestion
	healthCheckQuery := `ALERTS{alertstate="firing",severity="critical",namespace=~"^openshift.*|^kube.*|^default$",namespace!~"openshift-customer-monitoring|openshift-logging"` + icQuery + "}"
```

mrbarge:
Tested and working :+1: 
/lgtm

PR #140, opened 2020-09-28
addMetrics Error Handling - OSD-5014

himanshudogra:
### What type of PR is this?
_(bug)_

It fixes the error handling part for the addMetric function

https://issues.redhat.com/browse/OSD-5014

_Fixes #_

- [Done ] Tested latest changes against a cluster
- [Done ] Ran `make generate` command locally to validate code changes


cmd/manager/main.go
```diff
@@ -170,13 +172,13 @@ func main() {
...
 // the Prometheus operator
-func addMetrics(ctx context.Context, cfg *rest.Config) {
+func addMetrics(ctx context.Context, cfg *rest.Config) error {
 	// Get the namespace the operator is currently deployed in.
 	operatorNs, err := k8sutil.GetOperatorNamespace()
 	if err != nil {
 		if errors.Is(err, k8sutil.ErrRunLocal) {
 			log.Info("Skipping CR metrics server creation; not running in a cluster.")
-			return
+			return err
```
mrbarge, line 181: This should return nil instead of err. We still want to support being able to run the operator locally. So if the err is "ErrRunLocal", it's fine for the metrics service to not be created - meaning that this should return nil, so as not to cause the whole operator to quit.

deploy/operator.yaml
```diff
@@ -52,3 +52,8 @@ spec:
                   fieldPath: metadata.namespace
             - name: OPERATOR_NAME
               value: "managed-upgrade-operator"
+          livenessProbe:
+            tcpSocket:
+              port: 8686
+            initialDelaySeconds: 5
```
mrbarge, line 58: Maybe let's increase this to 10 or something. 5 seconds seemed to be cutting it pretty close during the tests we did during the meet.

himanshudogra: Done.

cmd/manager/main.go
himanshudogra: Certainly. corrected the changes.

cmd/manager/main.go
```diff
@@ -210,6 +212,7 @@ func addMetrics(ctx context.Context, cfg *rest.Config) {
 			log.Info("Install prometheus-operator in your cluster to create ServiceMonitor objects", "error", err.Error())
 		}
 	}
+	return err
```
mrbarge, line 215: Having tested this out, what we're seeing is that if the metrics service fails to create (which is normal and expected in the case of the pod restarting) it causes the operator to fail.

So likely what needs to happen is:
- if the `serveCRMetrics` call fails and returns an error, this should pass the error back up the chain and cause the operator to restart.
- everything else in the function is fine to fail, as there are legitimate reasons for them to fail, so this `return err` at the end is not needed.

cmd/manager/main.go
```diff
@@ -215,6 +217,7 @@ func addMetrics(ctx context.Context, cfg *rest.Config) error {
 	service, err := metrics.CreateMetricsService(ctx, cfg, servicePorts)
 	if err != nil {
 		log.Info("Could not create metrics Service", "error", err.Error())
+		return err
```
mrbarge, line 220: We'd only want this to return an error if the metrics service couldn't be created and the reason **wasn't** because it already existed.

So this would require some examination of the error returned by CreateMetricsService() to see what the error looks like when the service already exists. Having had a quick glance at the CreateMetricsService code, this might be a little awkward, because it doesn't pass the k8s error up the chain, it wraps it in another error.

What I would recommend is running the operator on a cluster in debug mode (check out Eric's video - I'll send a link) and seeing what error is returned when the service already exists. Then writing a code block accordingly such that "if the error isn't like this, then return an error".

cmd/manager/main.go
```diff
@@ -229,9 +232,11 @@ func addMetrics(ctx context.Context, cfg *rest.Config) error {
 		// ErrServiceMonitorNotPresent, which can be used to safely skip ServiceMonitor creation.
 		if err == metrics.ErrServiceMonitorNotPresent {
 			log.Info("Install prometheus-operator in your cluster to create ServiceMonitor objects", "error", err.Error())
+			return err
```
mrbarge, line 235: Same here as in my earlier comment - we would only want to error if it's a case that the service monitor couldn't be created and it's for a reason other than "it already exists".

bmeng:
I am not sure if we need to do this?
The functions were generated by the operator-sdk, we should trust it unless we have strong reason?

mrbarge:
> I am not sure if we need to do this?
> The functions were generated by the operator-sdk, we should trust it unless we have strong reason?

This is one case where I think the operator-sdk could be improved, which is ultimately where these sort of fixes should land. 

The primary thing I want to see out of this PR is that we are handling the case of the CR metrics server failing to create. For  everything else (eg. failure to create service monitors) we'd receive low priority alerts on anyway.

@himanshudogra - as it stands, maybe just remove the 'return err' on lines 220 and 235 because they will be disruptive.

himanshudogra:
> @himanshudogra - as it stands, maybe just remove the 'return err' on lines 220 and 235 because they will be disruptive.

Removed the err from highlighted lines. I am still under in the testing phase as per describing [here](https://github.com/openshift/managed-upgrade-operator/pull/140/#discussion_r503614952)

mrbarge:
I think this is safe to merge - I gave it a try on a cluster and it's behaving fine. Thanks @himanshudogra !

/lgtm

PR #141, opened 2020-09-30
Reuse node draining on scaled up nodes

sedroche:
### What type of PR is this?
bug

Adds the drain logic to our scaled up nodes. Troublesome pods can land on these nodes and our drain rules must be respected when scaling down too.
https://issues.redhat.com/browse/OSD-5248


mrbarge:
Don't want to hold this up. Will watch E2E overnight for validation.
/lgtm

PR #151, opened 2020-10-21
Stop crashloop for ocm error

sedroche:
### What type of PR is this?
bug

Seen on gitlab-webhook where the cluster has incorrect account setup in ocm.
The event manager does a call to ocm during the main.go startup routine which, if it errors, crashloops the entire pod.

_Fixes #_ https://issues.redhat.com/browse/OSD-5575

### DIFF
--- a/pkg/notifier/ocmnotifier.go
+++ b/pkg/notifier/ocmnotifier.go
@@ -43,14 +43,7 @@ func New(client client.Client, ocmBaseUrl *url.URL, upgradeConfigManager upgrade
 		},
 	}
 
-	// Get the cluster ID once rather than each time
-	cluster, err := getClusterFromOCMApi(client, httpClient, ocmBaseUrl)
-	if err != nil {
-		return nil, fmt.Errorf("Failed to retrieve external cluster ID")
-	}
-
 	return &ocmNotifier{
-		clusterID:            cluster.Id,
 		client:               client,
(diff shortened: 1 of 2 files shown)


pkg/notifier/ocmnotifier.go
```diff
@@ -73,12 +64,16 @@ type ocmNotifier struct {
 
 func (s *ocmNotifier) NotifyState(value NotifyState, description string) error {
 
-	policyId, err := s.getPolicyIdForUpgradeConfig()
+	clusterId, err := s.getInternalClusterId()
```
mrbarge, line 67: So the reason I initially put clusterID in the struct was to minimize the number of times a cluster has to reach out to OCM to get its cluster ID - do it once. This approach, if I'm reading it right, will result in a call every X minutes and makes the cluster very chatty with OCM.

Can we change this so that if it gets the cluster ID successfully once, it permanently caches it and doesn't attempt to re-get it (until the operator pod next restarts)?

mrbarge: Ignore me - this is fine.

mrbarge:
This looks fine. Ideally, would be good to have a test case where no cluster ID is returned when the request is made, but that's going to be real fiddly to do with the server mock right now (there's got to be a better way to do this then what currently exists). 

But for now, this tested okay and didn't blow up on my test cluster, so..

/lgtm

PR #159, opened 2020-10-26
Return next occuring of multiple upgrade policies

dofinn:
### What type of PR is this?
feature

This enables the operator to iterate through all available upgrade_policies and process the next occuring. 

[OSD-5564](https://issues.redhat.com/browse/OSD-5564)

### Further notes
This is vulnerable https://github.com/openshift/managed-upgrade-operator/blob/master/pkg/notifier/ocmnotifier.go#L107-L115. Its possible to create an upgrade policy of `schedule_type` = "automatic" or "manual" that have both the same `version` and `next_run`. Will need protection from this upstream. 

- [x] Tested latest changes against a cluster (validates correct policy is used and cleared with notifications)

### DIFF
--- a/pkg/ocmprovider/ocmprovider.go
+++ b/pkg/ocmprovider/ocmprovider.go
@@ -33,12 +33,11 @@ var log = logf.Log.WithName("ocm-config-getter")
 // Errors
 var (
 	ErrProviderUnavailable = fmt.Errorf("OCM Provider unavailable")
-	ErrClusterIdNotFound = fmt.Errorf("cluster ID can't be found")
+	ErrClusterIdNotFound   = fmt.Errorf("cluster ID can't be found")
 	ErrMissingChannelGroup = fmt.Errorf("channel group not returned or empty")
-	ErrRetrievingPolicies = fmt.Errorf("could not retrieve provider upgrade policies")
-	ErrProcessingPolicies = fmt.Errorf("could not process provider upgrade policies")
+	ErrRetrievingPolicies  = fmt.Errorf("could not retrieve provider upgrade policies")
+	ErrProcessingPolicies  = fmt.Errorf("could not process provider upgrade policies")
 )
-
 
 func New(client client.Client, ocmBaseUrl *url.URL) (*ocmProvider, error) {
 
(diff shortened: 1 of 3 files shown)


pkg/ocmprovider/ocmprovider.go
```diff
@@ -150,21 +149,33 @@ func (s *ocmProvider) Get() ([]upgradev1alpha1.UpgradeConfigSpec, error) {
...
 	}
 
-	// Apply the Upgrade Policies to the cluster
-	specs, err := buildUpgradeConfigSpecs(upgradePolicies, cluster, s.client)
-	if err != nil {
-		log.Error(err, "cannot build UpgradeConfigs from policy")
-		return nil, ErrProcessingPolicies
-	}
+	// Get the next occuring policy from the available policies
+	if len(upgradePolicies.Items) > 0 {
```
mrbarge, line 160: I haven't tested this to be sure, but if there are no policies returned, this will return nil and nil, which will break the upgradeconfigmanager. If nil is returned as the error, it expects a proper list of UpgradeConfigSpecs ([link](https://github.com/openshift/managed-upgrade-operator/blob/1f0d78c7b314c1d2008461823535b672349408c7/pkg/upgradeconfigmanager/upgradeconfigmanager.go#L184)). Can you please test this scenario and make sure it works ok?

mrbarge: (And if this is indeed an issue, it means we need a test for it, because evidently our existing ones didn't catch it :/)

mrbarge:
Other than what I noted earlier, I'm fine with this.

dofinn:
Upgrade completed, appropriate policy ID was removed upon completion

pkg/ocmprovider/ocmprovider.go
dofinn: This is actually a result of BDD already, I had to add the above to account for 0 upgradePolicies. Also it appears that an empty slice should be returned only when encoding to JSON, otherwise a `nil` slice is fine: https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

PR #162, opened 2020-10-27
Pin mockgen

sedroche:


ravitri:
Adding LGTM for now. For future, this is to be taken up in https://issues.redhat.com/browse/OSD-5673 to align go mods for tests/tools in a better way and not include as part of go.mod file

PR #166, opened 2020-11-03
compare upgradeconfig with upgradepolicy right before commence

bmeng:
### What type of PR is this?
feature

Fixes [OSD-5006](https://issues.redhat.com/browse/OSD-5006)

### DIFF
--- a/pkg/controller/upgradeconfig/upgradeconfig_controller.go
+++ b/pkg/controller/upgradeconfig/upgradeconfig_controller.go
@@ -2,17 +2,18 @@ package upgradeconfig
 
 import (
 	"context"
+	"time"
+
 	"github.com/openshift/managed-upgrade-operator/pkg/upgradeconfigmanager"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
-	"time"
 
 	"github.com/go-logr/logr"
 	"github.com/hashicorp/go-multierror"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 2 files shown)


pkg/controller/upgradeconfig/upgradeconfig_controller.go
```diff
@@ -188,6 +189,29 @@ func (r *ReconcileUpgradeConfig) Reconcile(request reconcile.Request) (reconcile
 			return reconcile.Result{}, err
 		}
 
+		upCoMan, err := upgradeconfigmanager.NewBuilder().NewManager(r.client)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		remoteSpecs, err := upCoMan.GetRemoteSpec()
```
mrbarge, line 196: Why not use the existing `Refresh()` here? If the current policy is different from the cluster's current UpgradeConfig, the local UpgradeConfig still should to be updated to whatever the new policy is.

`Refresh()` will do that already, and returns a boolean value to indicate if there was indeed a difference in policy vs UpgradeConfig, which also negates the need for the logic further below.

mrbarge: Another note - am thinking this should be going in within the `if schedulerResult.IsReady` block, ie. it should only be executed when we've reached the time to upgrade. Where it's placed right now will result in a call/refresh to OCM on each reconcile whilst the UC is in a pending state.

dofinn:
LGTM. Will let @mrbarge have final say :+1:

mrbarge:
Gave this a test today. Scenario was:

- Set up a policy
- Waited for cluster to sync policy
- Deleted policy
- Created new policy with a different time/version 2 hours in the future
- Waited for cluster to reach 'upgrade' time of first policy
- Cluster replaced the policy with the new policy
- Cluster reported its next upgrade as ~2 hours in the future as expected.

This seems good to go. I will lgtm as soon as the tests pass.

PR #168, opened 2020-11-06
Complete clusterversion client

sedroche:
### What type of PR is this?
refactor

Finished out the cv client

### DIFF
--- a/pkg/clusterversion/cv.go
+++ b/pkg/clusterversion/cv.go
@@ -2,18 +2,27 @@ package clusterversion
 
 import (
 	"context"
+	"fmt"
 
 	configv1 "github.com/openshift/api/config/v1"
-	upgradev1alpha1 "github.com/openshift/managed-upgrade-operator/pkg/apis/upgrade/v1alpha1"
-	"k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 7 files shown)


pkg/clusterversion/cv.go
```diff
@@ -41,18 +50,96 @@ func (cvb *clusterVersionClientBuilder) New(c client.Client) ClusterVersion {
...
+		return &HasDegradedOperatorsResult{
+			Degraded: []string{},
+		}, err
+	}
+
+	degradedOperators := []string{}
+	for _, co := range operatorList.Items {
+		for _, condition := range co.Status.Conditions {
+			if (condition.Type == configv1.OperatorDegraded && condition.Status == configv1.ConditionTrue) || (condition.Type == configv1.OperatorAvailable && condition.Status == configv1.ConditionFalse) {
+				degradedOperators = append(degradedOperators, co.Name)
```
mrbarge, line 122: Tiny one here (I know this is a pre-existing bug, I just only noticed it when testing this PR) - can we add a `break` after this line, otherwise in the case that an operator is both degraded and unavailable it adds an entry into the `degradedOperators` for each failing status condition, and it need only appear once.

mrbarge:
Reviewed and tested this - looks good to me, I'm happy to lgtm this once you've responded to my 1 small comment.

PR #181, opened 2020-12-04
Delete and re-create UpgradeConfig on spec change OSD-5970

dofinn:
### What type of PR is this?
bug

This resolves a racey condition where a new upgrade_policy was updating the current UpgradeConfig without deleting it. If the UpgradeConfig was in a failed state, the new upgrade_policy inherited that state rendering MUO unable to proceed with the new upgradepolicy specs updated into the current UpgradeConfig CR
https://issues.redhat.com/browse/OSD-5970 

Tested by deleting and re-creating upgrade policies that only differ in upgrade_start time as that is where the original condition was evident.

- [X] Tested latest changes against a cluster
- [X] Ran `make generate` command locally to validate code changes

### DIFF
--- a/pkg/upgradeconfigmanager/upgradeconfigmanager.go
+++ b/pkg/upgradeconfigmanager/upgradeconfigmanager.go
@@ -269,7 +269,7 @@ func (s *upgradeConfigManager) Refresh() (bool, error) {
 
 			// Confirm object is deleted prior to creating
 			if !confirmDeletedUpgrade {
-				wait.PollImmediate(5*time.Second, 60*time.Second, func() (bool, error) {
+				err = wait.PollImmediate(5*time.Second, 60*time.Second, func() (bool, error) {
 					_, err := s.Get()
 					if err != nil {
 						if err == ErrUpgradeConfigNotFound {
@@ -281,6 +281,9 @@ func (s *upgradeConfigManager) Refresh() (bool, error) {
 					}
 					return false, nil
 				})
+				if err != nil {
+					return false, fmt.Errorf("Unable to confirm deletion of current UpgradeConfig")
+				}
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


pkg/upgradeconfigmanager/upgradeconfigmanager.go
```diff
@@ -245,18 +246,22 @@ func (s *upgradeConfigManager) Refresh() (bool, error) {
...
-		log.Info("cluster upgrade spec has changed, will update")
-		err = s.client.Update(context.TODO(), &replacementUpgradeConfig)
-		if err != nil {
-			if errors.IsNotFound(err) {
-				// couldn't update because it didn't exist - create it instead.
-				err = s.client.Create(context.TODO(), &replacementUpgradeConfig)
+		if foundUpgradeConfig {
+			// Delete and re-create the resource
+			log.Info("cluster upgrade spec has changed, will delete and re-create.")
+			err = s.client.Delete(context.TODO(), currentUpgradeConfig)
```
sedroche, line 252: Can you set `--grace-period` and `--force` in the options? I have found deletion can be tricky as it marks a resource for deletion but can take some time.
You might hit errors on recreation if its not gone

mrbarge: Yeah that's a good point. 
I think looking at `DeletePropagationForeground` is the preferred way here. https://github.com/ericchiang/k8s/blob/08b7bf46703ad24b5420601bf6df6d69b4b74201/resource.go#L72

I'm also not sure if we could potentially face resource version conflict errors - because this operates in a separate goroutine to the UC controller, you could potential have a situation where the UC controller updates the UpgradeConfig in-between the time that the UC Manager reads the UpgradeConfig and deletes it. (but again, not sure if deletes are subject to this)

dofinn: Resolved w/ https://github.com/openshift/managed-upgrade-operator/pull/181/commits/44f0d1fe3d7e0af222b2e8dd86fd26ba635f4a63

mrbarge:
Have given this a bunch of tests and think it's behaving fine.
/lgtm

PR #183, opened 2020-12-11
OSD-5724 Support external dep availability checks

dofinn:
### What type of PR is this?
feature

Required to validate dependent services are available prior to upgrading. 

https://issues.redhat.com/browse/OSD-5724

This PR introduces an AvailabilityChecker interface and support HTTP implementation. The interface itself only requires AvailabilityCheck() to implement. Within the upgrader.go code, the interface is passed to each upgrade step as a slice so that muliple implementations can/could be passed at once. In future we could expand this to include DNS, registry checks. 

This PR should no be merged until required config is merged in MCC. 

- [] Included documentation changes with PR


pkg/availabilitychecks/http.go
```diff
@@ -0,0 +1,102 @@
...
+
+type HTTPConfig struct {
+	Targets []string
+	Timeout time.Duration
+}
+
+// NewHTTPAvailabilityChecker accepts a Config and attempts to ascert its type
+// as HTTPConfig. If ascertation is true a HTTPAvailabilityChecker is returned.
+// Else, an error is returned advising failed ascertation.
+//func NewHTTPAvailabilityChecker(c Config) (*HTTPAvailabilityChecker, error) {
```
mrbarge, line 23: Can be removed?

dofinn:
Pending https://github.com/openshift/managed-cluster-config/pull/596

pkg/availabilitychecks/http.go
dofinn: https://github.com/openshift/managed-upgrade-operator/pull/183/commits/c6352507a872fce7e840f6a93e39ffad2ee76aa9

pkg/osd_cluster_upgrader/upgrader.go
```diff
@@ -167,8 +177,38 @@ func EnsureExtraUpgradeWorkers(c client.Client, cfg *osdUpgradeConfig, s scaler.
...
+	for _, check := range availabilityCheckers {
+		logger.Info(fmt.Sprintf("Checking availability check for %T", check))
+		err := check.AvailabilityCheck()
+		if err != nil {
+			logger.Info(fmt.Sprintf("Failed availability check for %T", check))
+			return false, err
+		}
+		logger.Info(fmt.Sprintf("Availability check complete for %T", check))
+	}
+	logger.Info("Returning false so upgrade does not commence")
```
mrbarge, line 206: Is this log correct?

pkg/availabilitychecks/retry.go
```diff
@@ -0,0 +1,32 @@
...
+
+func init() {
+	rand.Seed(time.Now().UnixNano())
+}
+
+type stop struct {
+	error
+}
+
+func retry(attempts int, sleep time.Duration, f func() error) error {
```
mrbarge, line 16: Up to you if you want to adopt it or not, not going to hold up a PR for it, but I use `backoff` [0] for a similar retry/backoff-with-jitter pattern for the upgradeconfigmanager and it's quite nice.
https://github.com/jpillora/backoff

pkg/osd_cluster_upgrader/upgrader.go
dofinn: https://github.com/openshift/managed-upgrade-operator/pull/183/commits/13eebbeb07f6c64f334ff62fb9cfdf4963235b25

pkg/osd_cluster_upgrader/config.go
```diff
@@ -85,6 +87,9 @@ func (cfg *osdUpgradeConfig) IsValid() error {
 	if cfg.UpgradeWindow.TimeOut < 0 {
 		return fmt.Errorf("Config upgrade window time out is invalid")
 	}
+	if cfg.ExtDependencyAvailabilityCheck.HTTP.Timeout <= 0 || cfg.ExtDependencyAvailabilityCheck.HTTP.Timeout > 60 {
+		return fmt.Errorf("Config HTTP timeout is invalid (Requires int between 1 - 60 inclusive)")
+	}
```
mrbarge, line 92: For robustness, I'd propose that we change this to only check it if there are actually HTTP availability URLs configured. That way, we can potentially leave out the whole `extDependencyAvailabilityChecks` from the configmap if desired, and it won't cause the controller to crashloop (its current behaviour in that scenario).

So something like:
```suggestion
	if len(cfg.ExtDependencyAvailabilityCheck.HTTP.URLS) > 0 && cfg.ExtDependencyAvailabilityCheck.HTTP.Timeout <= 0 || cfg.ExtDependencyAvailabilityCheck.HTTP.Timeout > 60 {
		return fmt.Errorf("Config HTTP timeout is invalid (Requires int between 1 - 60 inclusive)")
	}
```

pkg/osd_cluster_upgrader/upgrader.go
```diff
@@ -167,8 +177,37 @@ func EnsureExtraUpgradeWorkers(c client.Client, cfg *osdUpgradeConfig, s scaler.
...
 
+// ExternalDependencyAvailabilityCheck validates that external dependencies of the upgrade are available.
+func ExternalDependencyAvailabilityCheck(c client.Client, cfg *osdUpgradeConfig, s scaler.Scaler, dsb drain.NodeDrainStrategyBuilder, metricsClient metrics.Metrics, m maintenance.Maintenance, cvClient cv.ClusterVersion, nc eventmanager.EventManager, upgradeConfig *upgradev1alpha1.UpgradeConfig, machinery machinery.Machinery, availabilityCheckers []ac.AvailabilityChecker, logger logr.Logger) (bool, error) {
+	upgradeCommenced, err := cvClient.HasUpgradeCommenced(upgradeConfig)
+	if err != nil {
+		return false, err
+	}
+	desired := upgradeConfig.Spec.Desired
+	if upgradeCommenced {
+		logger.Info(fmt.Sprintf("ClusterVersion is already set to Channel %s Version %s, skipping %s", desired.Channel, desired.Version, upgradev1alpha1.UpgradeScaleUpExtraNodes))
```
mrbarge, line 188:
```suggestion
		logger.Info(fmt.Sprintf("ClusterVersion is already set to Channel %s Version %s, skipping %s", desired.Channel, desired.Version, upgradev1alpha1.ExtDepAvailabilityCheck))
```

pkg/osd_cluster_upgrader/upgrader.go
```diff
@@ -167,8 +177,37 @@ func EnsureExtraUpgradeWorkers(c client.Client, cfg *osdUpgradeConfig, s scaler.
...
+
+	for _, check := range availabilityCheckers {
+		logger.Info(fmt.Sprintf("Checking availability check for %T", check))
+		err := check.AvailabilityCheck()
+		if err != nil {
+			logger.Info(fmt.Sprintf("Failed availability check for %T", check))
+			return false, err
+		}
+		logger.Info(fmt.Sprintf("Availability check complete for %T", check))
+	}
```
mrbarge, line 205: Substituting `%v` for `%T` here gives more useful and actionable info.

eg. it results in
```
Checking availability check for &{[https://quay.io/health https://apiasdfasdf.stage.openshift.com] 10s}
```

instead of:
```
Checking availability check for *availabilitychecks.HTTPAvailabilityChecker
```

pkg/osd_cluster_upgrader/upgrader_test.go
```diff
@@ -288,19 +292,44 @@ var _ = Describe("ClusterUpgrader", func() {
...
+			gomock.InOrder(
+				mockCVClient.EXPECT().HasUpgradeCommenced(gomock.Any()).Return(false, nil),
+				mockAC.EXPECT().AvailabilityCheck().Return(fakeErr),
+			)
+
+			result, err := ExternalDependencyAvailabilityCheck(mockKubeClient, config, mockScalerClient, mockDrainStrategyBuilder, mockMetricsClient, mockMaintClient, mockCVClient, mockEMClient, upgradeConfig, mockMachineryClient, []ac.AvailabilityChecker{mockAC}, logger)
+			Expect(err).To(HaveOccurred())
+			Expect(result).To(BeFalse())
+		})
+
```
mrbarge, line 324: Could also add a test to ensure that availability checking doesn't occur when a cluster is upgrading.

```suggestion
		It("will not perform availability checking if the cluster is upgrading", func() {
			gomock.InOrder(
				mockCVClient.EXPECT().HasUpgradeCommenced(gomock.Any()).Return(true, nil),
			)
			result, err := ExternalDependencyAvailabilityCheck(mockKubeClient, config, mockScalerClient, mockDrainStrategyBuilder, mockMetricsClient, mockMaintClient, mockCVClient, mockEMClient, upgradeConfig, mockMachineryClient, []ac.AvailabilityChecker{mockAC}, logger)
			Expect(err).NotTo(HaveOccurred())
			Expect(result).To(BeTrue())
		})
```

pkg/osd_cluster_upgrader/upgrader.go
dofinn: I wanted the implementation to be raised in the logs for future cases where multiple implementations may exist. The failed url/s are returned to the caller via `err`

pkg/availabilitychecks/retry.go
dofinn: i would like to revist this during a refactoring card.

pkg/osd_cluster_upgrader/upgrader_test.go
dofinn: Thanks: https://github.com/openshift/managed-upgrade-operator/pull/183/commits/bff179fe00a8e1fa2b77d3737f6df5dc9faead8f

PR #184, opened 2020-12-15
liveness check for MUO

dofinn:
### What type of PR is this?
feature

Currently MUO has no liveness checks

Nil

- [n/a] Included documentation changes with PR

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -34,6 +34,13 @@ spec:
           command:
           - managed-upgrade-operator
           imagePullPolicy: Always
+          livenessProbe:
+            exec:
+              command:
+                - sh
+                - -c
+                - OK=$(curl http://localhost:8686/healthz); if [ "$OK" != "ok"
+                  ]; then exit 1;fi
           env:
             - name: WATCH_NAMESPACE
               valueFrom:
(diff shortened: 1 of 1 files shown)


deploy/operator.yaml
```diff
@@ -34,6 +34,13 @@ spec:
           command:
           - managed-upgrade-operator
           imagePullPolicy: Always
+          livenessProbe:
+            exec:
+              command:
+                - sh
+                - -c
+                - OK=$(curl http://localhost:8686/healthz); if [ "$OK" != "ok"
+                  ]; then exit 1;fi
```
mrbarge, line 43: I swear we covered this in #140 , apparently not. Anyway, this is better done as:

```suggestion
            httpGet:
              path: /healthz
              port: 8686
```

You'll see in the comments of #140 that it should probably have a generous `initialDelaySeconds` value because the local metrics service doesn't tend to start up immediately.

dofinn: this uses the pods IP of which the /healthz is not accessible.

dofinn: resolved: https://github.com/openshift/managed-upgrade-operator/pull/184/commits/e0bf41c24154a75318358b88a5df804fffed5d5a

PR #186, opened 2021-01-14
Reset metrics when upgrade failed OSD-6094

dofinn:
Created reset functions for metrics that are triggered prior to an
upgrade being cancelled due to upgrade window timeout

### What type of PR is this?
bug

Correctly resets metrics after a failed upgrade

[OSD-6094](https://issues.redhat.com/browse/OSD-6094)


mrbarge:
/hold 
just whilst we phase our promotions.
Will unhold after #184 is promoted.

bmeng:
https://github.com/openshift/managed-upgrade-operator/blob/master/pkg/metrics/metrics.go#L416
We have the function to reset all the metrics.
Maybe we can also use this instead of separated resets?

PR #195, opened 2021-02-04
add resource limit/request for the operator

bmeng:
set the resource limits/requests for the operator

based on the usage for a running cluster, the cpu usage will be around 2.5m and the memory usage will be around 50M
Set the requests so that the pod will be scheduled only when the node has enough resources
and also the limits in case the operator eating up the resources

- [ ] Ran `make generate` command locally to validate code changes

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -34,6 +34,13 @@ spec:
           command:
           - managed-upgrade-operator
           imagePullPolicy: Always
+          resources:
+            requests:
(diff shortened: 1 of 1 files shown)


bmeng:
cpu usage including an upgrade
cpu

memory usage including an upgrade
memory

bmeng:
hmm, the cpu screenshot is for catalog pod, the MUO pod is the line below that. it is even less

PR #199, opened 2021-03-10
changed sop link

a7vicky:
I added updated sop link for the distribution of `UpgradeConfig` CR.


a7vicky:
Opened a new one #200

PR #200, opened 2021-03-10
fix doc link.

a7vicky:
I have added new sop doc link for `UpgradeConfig` CR creation and distribution details.


mrbarge:
Thanks for spotting these!

/lgtm
/approve

PR #202, opened 2021-03-12
OSD-6718 pc

dofinn:
### What type of PR is this?
feature/bug

We dont set priority class

[osd-6718](https://issues.redhat.com/browse/osd-6718)

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -22,6 +22,7 @@ spec:
               - key: node-role.kubernetes.io/infra
                 operator: Exists
             weight: 1
+      priorityClassName: "system-cluster-critical"
       tolerations:
         - operator: Exists
           key: node-role.kubernetes.io/infra


deploy/operator.yaml
```diff
@@ -22,6 +22,7 @@ spec:
               - key: node-role.kubernetes.io/infra
                 operator: Exists
             weight: 1
+      priorityClassName: "system-cluster-critical"
```
ravitri, line 25: From a point of view, I don't think the MUO pod is "critical" for the Openshift cluster to function. The cluster would function just the same even without MUO pod being scheduled. The existing clusteroperators can have such a value because mostly they are actually critical for the cluster to operate. I think the investigation around https://issues.redhat.com/browse/OSD-6717 could be given more thought first and then we can go ahead with this change.

dofinn: It is critical for an OSD upgrade, if the MUO can not be scheduled because a RHOAM pod has greater priority then the managed-upgrade will fail.

ravitri: Thanks Dom! My query/concern here was different though. I understand that MUO pod should definitely have high priority since upgrades are vital for the cluster. Question here was about if MUO pod should have 2nd highest cluster priority of `system-cluster-critical` level priority or something lesser than that but high enough that customer/app (Eg. RHOAM) pods don't preempt it? I think that any app pods should have lesser priority configured than the "lowest" value of any priorityClass which is available in Openshift by default.  The lowest is `openshift-user-critical` on a 4.7 cluster for now with value 1000000000. But I see that RHOAM had the same value set for their pods and that's seems to be the concern here IMO. As an example, the ClusterLogging operator defines a priority value of 1000000 by default and that seems a better value for an app addon to have comparatively.

Am referring to review a possible scenario such as https://github.com/openshift/cluster-monitoring-operator/pull/1062 where the `system-cluster-critical` was infact problematic for CMO and it's priority had to be lowered to `openshift-user-critical` priority for the `monitoring` clusteroperator so that the pod is allowed too OOM kill and not starve the node itself. Though that scenario was specific to memory consumption issue of Prometheus and is unlikely for MUO to face AFAIK, I thought it's better to review the priorities and assign a priorityClass (or create new one) based on what level the pod actually needs than allocate a system level priorityClass just because it's amongst the highest and available.

For now, even `system-cluster-critical` would work ofcourse for MUO pod so lgtm from my end for now but as per https://issues.redhat.com/browse/OSD-6717 we can perhaps revisit this point again in better detail for MUO and other SRE operators.

dofinn: Thanks for airing this. Not sure I agree with the approach to lower monitoring as per above. Will do some digging.

PR #214, opened 2021-03-31
handle the metrics and the extra machine set when upgrade failed

bmeng:
### What type of PR is this?
bug/enhancement

force scale down the extra machine when upgrade failed
reset metric when node scale failed

_Fixes [OSD-5723](https://issues.redhat.com/browse/OSD-5723) OSD-6094_

### DIFF
--- a/pkg/controller/nodekeeper/nodekeeper_controller.go
+++ b/pkg/controller/nodekeeper/nodekeeper_controller.go
@@ -3,13 +3,14 @@ package nodekeeper
 import (
 	"context"
 	"fmt"
-	"github.com/openshift/managed-upgrade-operator/util"
 	"time"
+
+	"github.com/openshift/managed-upgrade-operator/util"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
-	"sigs.k8s.io/controller-runtime"
+	controllerruntime "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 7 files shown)


mrbarge:
/hold 
for promotion synchronisation

pkg/osd_cluster_upgrader/upgrader.go
```diff
@@ -641,9 +641,15 @@ func performUpgradeFailure(metricsClient metrics.Metrics, nc eventmanager.EventM
 	// flag window breached metric
 	metricsClient.UpdateMetricUpgradeWindowBreached(upgradeConfig.Name)
 
+	// TearDown the extra machineset
```
mrbarge, line 644: I think it may be better if this happens at the start of the function. That way, if the scale down were to somehow fail and retry on the next reconcile, it isn't attempting to resend failure notifications (which immediately triggers a policy cancellation/deletion on the OCM end) and overwriting the window breach metric. The successful scale-down should be a prerequisite before the notification of failure to OCM

mrbarge:
I gave this a test with a real upgrade:

- Let the upgrade progress through to scale up nodes
- Whilst scale-up was occuring, I artifically induced upgrade failure by making the external dependency check step now permanently fail.
- The upgrade failed after the window timeout elapsed.
- I verified the upgrade scaled worker was deleted as part of the upgrade fail.

Other than the separate comment I've made about re-ordering the code in the `performUpgradeFailure`, this looks and tests good to me - great change.

dofinn:
:lgtm

PR #219, opened 2021-04-14
OSD-6778 MUO Refactor metrics

dofinn:
### What type of PR is this?
refactor

Refactors metrics by writing state to the status fields of the UpgradeConfig and using a prometheus collector to then review the UpgradeConfig and collect/report metrics based on that state. 

This means all previous metrics (bar providerSync failure) have been moved to a status within the active `UpgradeConfig.Status.History[0]`

https://issues.redhat.com/browse/OSD-6778
#[OSD-6778](https://issues.redhat.com/browse/OSD-6778)

---

Changes to boilerplate were to enable 

```
./boilerplate/_lib/container-make generate
```

Unsure if these are being backported. 

---

Testing this will require you to build your own image and update the CSV to pull it. i

`ALLOW_DIRTY_CHECKOUT=true make docker-build; podman push quay.io/$(whoami)/managed-upgrade-operator:latest`

Update CSV to pull latest.

---

Provider Sync failed will need to be reviewed as its a chick and egg situation where we can not write a status to an UpgradeConfig that does not exist.

---

Follow ups:

- update metrics docs

- [o] Included documentation changes with PR


dofinn:
20210414_124137_grim

dofinn:
will need to merge MCC first to alert on new metrics

pkg/upgraders/osd/upgrader.go
```diff
@@ -142,11 +143,19 @@ func PreClusterHealthCheck(c client.Client, cfg *osdUpgradeConfig, scaler scaler
 
 	ok, err := performClusterHealthCheck(c, metricsClient, cvClient, cfg, logger)
 	if err != nil || !ok {
-		metricsClient.UpdateMetricClusterCheckFailed(upgradeConfig.Name)
+		upgradeConfig.Status.History[0].HealthCheck.Failed = true
```
sedroche, line 146: imo everything in upgraders should be a condition. (returned from the upgrader)
These statuses will change between upgraders so we can't define them on the CRD

sedroche: Actually, everything outside the controllers should be

pkg/controller/upgradeconfig/upgradeconfig_controller.go
```diff
@@ -261,6 +271,13 @@ func (r *ReconcileUpgradeConfig) Reconcile(request reconcile.Request) (reconcile
 		reqLogger.Info("Cluster is already upgraded")
 		return reconcile.Result{}, nil
 	case upgradev1alpha1.UpgradePhaseFailed:
+		instance.Status.History[0].WindowBreached = true
```
sedroche, line 274: `[0]` makes sense here but its best to retrieve the history based on the version

deploy/crds/upgrade.managed.openshift.io_upgradeconfigs_crd.yaml
```diff
@@ -157,22 +185,49 @@ spec:
...
                     format: date-time
                     type: string
+                  workerTimeout:
+                    description: A boolean advising if the clusters control plane upgrade has timed out.
+                    type: boolean
                 required:
                   - phase
                 type: object
               type: array
+            notificationEventSent:
```
sedroche, line 221: Not hugely familiar with this part of the code anymore but notifications on the status seems wrong to me

cmd/manager/main.go
```diff
@@ -175,6 +177,15 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Setup operator specific metrics
+	muoMetrics := prometheus.NewRegistry()
+	muoMetrics.MustRegister(
+		prometheus.NewGoCollector(),
+		prometheus.NewProcessCollector(prometheus.ProcessCollectorOpts{
+			Namespace: "muo",
```
mrbarge, line 185: Think it's worth being explicit here, `managed-upgrade-operator`.

deploy/crds/upgrade.managed.openshift.io_upgradeconfigs_crd.yaml
```diff
@@ -148,6 +154,28 @@ spec:
                         - type
                       type: object
                     type: array
+                  controlPlaneTimeout:
+                    description: A boolean advising if the clusters control plane upgrade has timed out.
+                    type: boolean
+                  healthCheck:
+                    description: A boolean advising if the pre upgrade health check failed.
```
mrbarge, line 161:
```suggestion
                    description: A boolean advising if the pre or post upgrade health check failed.
```

deploy/crds/upgrade.managed.openshift.io_upgradeconfigs_crd.yaml
```diff
@@ -148,6 +154,28 @@ spec:
...
+                      failed:
+                        type: boolean
+                      state:
+                        enum:
+                          - pre_upgrade
+                          - post_upgrade
+                        type: string
+                    type: object
+                  nodeDrain:
+                    description: A boolean advising if the UpgradeConfig provider sync failed.
```
mrbarge, line 172:
```suggestion
                    description: A boolean advising if an upgrade node drain has failed.
```

deploy/crds/upgrade.managed.openshift.io_upgradeconfigs_crd.yaml
```diff
@@ -157,22 +185,49 @@ spec:
...
+                    description: A boolean advising if the UpgradeConfig window has been breached.
+                    type: boolean
                   workerCompleteTime:
                     format: date-time
                     type: string
                   workerStartTime:
                     format: date-time
                     type: string
+                  workerTimeout:
+                    description: A boolean advising if the clusters control plane upgrade has timed out.
```
mrbarge, line 215:
```suggestion
                    description: A boolean advising if an upgrading worker has taken too long to upgrade.
```

deploy/crds/upgrade.managed.openshift.io_upgradeconfigs_crd.yaml
mrbarge: I get what you mean. I'm kind of 50/50 on its incorporation into the Status. 

I tend to lean towards their inclusion here as being fine. On the one hand notifications are currently a core, product-agnostic aspect of the operator's behaviour. On the other hand, I regularly think they shouldn't be.
On the other-other hand, the notifications are an integral, standalone step of the OSD upgrader, which is why their representation here carries some weight.

Keen to engage in discussion about it though.

pkg/controller/upgradeconfig/upgradeconfig_controller.go
mrbarge: +1 to this. All lookups in History should probably be done with `GetHistory(<version>)` and a `nil` check performed afterwards to be safe, in the interests of accidental panic prevention. 

If that starts to make the code look ugly and excessively verbose, perhaps it could use wrapping in some function that performs the required updates safely.

pkg/upgraders/osd/upgrader.go
mrbarge: Yeah.. agree that this makes sense (upon reflection of the related comment about Notifications in the CRD).

We should ideally want upgrade steps to be flexible and adaptable and not necessarily have any guarantees of them needing to be present, having them explicitly modelled in the CRD goes against that somewhat.

pkg/apis/upgrade/v1alpha1/upgradeconfig_types.go
```diff
@@ -63,6 +96,34 @@ type UpgradeHistory struct {
 	WorkerStartTime *metav1.Time `json:"workerStartTime,omitempty"`
 
 	WorkerCompleteTime *metav1.Time `json:"workerCompleteTime,omitempty"`
+
+	// A boolean advising if the pre upgrade health check failed.
```
mrbarge, line 100:
```suggestion
	// A boolean advising if the pre or post upgrade health check failed.
```

pkg/apis/upgrade/v1alpha1/upgradeconfig_types.go
```diff
@@ -63,6 +96,34 @@ type UpgradeHistory struct {
...
+
+	// A boolean advising if the clusters control plane upgrade has timed out.
+	// +kubebuilder:validation:Optional
+	ControlPlaneTimeout bool `json:"controlPlaneTimeout"`
+
+	// A boolean advising if the clusters control plane upgrade has timed out.
+	// +kubebuilder:validation:Optional
+	WorkerTimeout bool `json:"workerTimeout"`
+
+	// A boolean advising if the UpgradeConfig provider sync failed.
```
mrbarge, line 120:
```suggestion
	// A boolean advising if an upgrading worker is failing to drain
```

pkg/apis/upgrade/v1alpha1/upgradeconfig_types.go
```diff
@@ -63,6 +96,34 @@ type UpgradeHistory struct {
...
+
+	// A boolean advising if the cluster verification has failed.
+	// +kubebuilder:validation:Optional
+	ClusterVerificationFailed bool `json:"clusterVerificationFailed,omitempty"`
+
+	// A boolean advising if the clusters control plane upgrade has timed out.
+	// +kubebuilder:validation:Optional
+	ControlPlaneTimeout bool `json:"controlPlaneTimeout"`
+
+	// A boolean advising if the clusters control plane upgrade has timed out.
```
mrbarge, line 116:
```suggestion
	// A boolean advising if the cluster's upgrading workers have not upgraded in time.
```

deploy/crds/upgrade.managed.openshift.io_upgradeconfigs_crd.yaml
sedroche: I'm 100% against it being on the status. Recorded in a status condition would be an easy generic way of recording this.

Other options are to maybe have the notifications manage a configmap for state and remove from the upgrader altogether (again, been a while since I looked in there)

pkg/collector/collector.go
```diff
@@ -0,0 +1,534 @@
...
+
+	controlPlaneTimeout := upgradeConfig.Status.History[0].ControlPlaneTimeout
+
+	if controlPlaneTimeout {
+		ch <- prometheus.MustNewConstMetric(
+			uc.coreUpgradeMetrics.controlPlaneTimeout,
+			prometheus.GaugeValue,
+			float64(1),
+			upgradeConfig.ObjectMeta.Name,
+			valueControlPlaneStarted,
```
mrbarge, line 498: Is there a reasoning for including the "control_plane_started" label in the metric related to timeout? (same goes for its inclusion below)

pkg/collector/collector.go
```diff
@@ -0,0 +1,534 @@
...
+	ch <- uc.coreUpgradeMetrics.configInvalid
+	ch <- uc.coreUpgradeMetrics.upgradeHealthCheckFailed
+	ch <- uc.coreUpgradeMetrics.scalingFailed
+	ch <- uc.coreUpgradeMetrics.clusterVerificationFailed
+	ch <- uc.coreUpgradeMetrics.controlPlaneTimeout
+	ch <- uc.coreUpgradeMetrics.workerTimeout
+	ch <- uc.coreUpgradeMetrics.nodeDrainFailed
+	ch <- uc.coreUpgradeMetrics.upgradeWindowBreached
+
+	ch <- uc.osdUpgradeMetrics.notificationEventSentFailed
```
mrbarge, line 180: Not really sure if you want to tackle it in this PR or wait until M7 but I feel there is lots of opportunity here to start thinking in advance about how this works when we have different upgraders around.

There are some early provisions in this PR for ARO collectors, but none of the code yet is looking at the upgrade type when collecting and sending various metrics.

ie. how will this function and (more specifically) the associated `collectUpgradeMetrics` behave when there are ARO and OSD upgraders?

ARO's metrics gathering doesn't need to be solved yet, but in terms of this PR I guess what I'm thinking is.. can some of these functions be changed in advance of ARO so that we only collect and send OSD related metrics if the upgradeconfig's type actually is "OSD"? (forgive me if that is actually happening and I've missed it). Solving that is half the battle of this issue and can be done ahead of time.

deploy/crds/upgrade.managed.openshift.io_upgradeconfigs_crd.yaml
mrbarge: Yeah, I should've seen your other note about status conditions before replying here. :) I agree on that.

boilerplate/_lib/container-make
```diff
@@ -14,7 +14,7 @@ CONTAINER_ENGINE=$(command -v podman || command -v docker)
 
 # First set up a detached container with the repo mounted.
 banner "Starting the container"
-container_id=$($CONTAINER_ENGINE run -d -v "$REPO_ROOT":"$REPO_ROOT" $IMAGE_PULL_PATH tail -f /dev/null)
+container_id=$($CONTAINER_ENGINE run --userns keep-id -d -v "$REPO_ROOT":"$REPO_ROOT":Z $IMAGE_PULL_PATH tail -f /dev/null)
```
mrbarge, line 17: This should be excluded from the MR and managed via `make boilerplate-update`.
Which we'll be likely need to do in the coming week or two.

deploy/crds/upgrade.managed.openshift.io_upgradeconfigs_crd.yaml
sedroche: :+1: Sent this between notifications too :smile:

boilerplate/_lib/container-make
dofinn: https://github.com/openshift/managed-upgrade-operator/pull/219/commits/16baf6e27566dac0d73a8b3c85eadeda77025f81

cmd/manager/main.go
dofinn: so i originally had the MetricsNamespace of "managed_upgrade" as that is the service the metrics are reporting. However for this collector, its more about the operator underneath. The reason I used an acronym was due to reviewing https://github.com/OpenObservability/OpenMetrics/blob/feb021c735e9d61ee5779f6e581b07f098304f9b/specification/OpenMetrics.txt#L1835-L1840

Happy to stand corrected tho.

pkg/collector/collector.go
```diff
@@ -0,0 +1,534 @@
...
+			}, nil),
+		upgradeWindowBreached: prometheus.NewDesc(
+			prometheus.BuildFQName(MetricsNamespace, subSystemUpgrade, "window_breached"),
+			helpUpgradeWindowBreached,
+			[]string{
+				keyUpgradeConfigName,
+			}, nil),
+	}
+
+	//	osdMetrics := &OSDUpgradeMetrics{
```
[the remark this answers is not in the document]
dofinn: This is a placeholder for follow up card. currently this metric still relies on old implemenatation, need to discuss best way forward.

deploy/crds/upgrade.managed.openshift.io_upgradeconfigs_crd.yaml
dofinn: Is there a way to make this field on visually present when .type= "OSD" ? This is kind of what i was thinking in regards to https://github.com/openshift/managed-upgrade-operator/pull/219/files#diff-82dc411a0ed86675e9efddaa43dadc4fc6a75ea0e672400429e8ea39424da44cR47-R68

pkg/collector/collector.go
```diff
@@ -0,0 +1,534 @@
...
+				keyUpgradeConfigName,
+			}, nil),
+		upgradeHealthCheckFailed: prometheus.NewDesc(
+			prometheus.BuildFQName(MetricsNamespace, subSystemCluster, "health_check_failed"),
+			helpUpgradeHealthCheckFailed,
+			[]string{
+				keyUpgradeConfigName,
+				keyState,
+			}, nil),
+		scalingFailed: prometheus.NewDesc(
```
sedroche, line 98: Can these upgrader metrics be moved to an osd upgrader collector?

cmd/manager/main.go
mrbarge: That's cool, I wasn't entirely sure what the implications where as to what happens with a namespace like that.
My main reason for noting it was that "muo" means nothing to anyone except us, so it might inhibit its potential usefulness to others who see it and have no context for what it means.
I kind of like `managed_upgrade_`.

pkg/collector/collector.go
mrbarge: This is basically the same thing I'm hoping to see with [my comment](https://github.com/openshift/managed-upgrade-operator/pull/219#discussion_r613004239). Would prefer we establish a clear separation of upgraders early on with this.

sedroche: :+1: Awesome

pkg/upgraders/osd/upgrader.go
```diff
@@ -635,6 +721,17 @@ func performUpgradeFailure(c client.Client, metricsClient metrics.Metrics, s sca
 	_, err := s.EnsureScaleDownNodes(c, nil, logger)
 	if err != nil {
 		logger.Error(err, "Failed to scale down the temporary upgrade machine when upgrade failed")
+		upgradeConfig.Status.History[0].Scaling.Failed = true
+		upgradeConfig.Status.History[0].Scaling.Dimension = "down"
+		if err = c.Status().Update(context.TODO(), upgradeConfig); err != nil {
+			return err
+		}
+		return err
```
mrbarge, line 729: Seems like this could just be shortened to
```suggestion
		return c.Status().Update(context.TODO(), upgradeConfig)
```

PR #225, opened 2021-04-22
add localprovider for upgradeconfigmanager

bmeng:
### What type of PR is this?
_feature_

_Fixes #OSD-6916_

### DIFF
--- a/pkg/localprovider/config.go
+++ b/pkg/localprovider/config.go
@@ -1,0 +1,20 @@
+package localprovider
+
+import "fmt"
+
+type LocalProviderConfig struct {
+	ConfigManager ConfigManager `yaml:"configManager"`
+}
+
+type ConfigManager struct {
+	LocalConfigName string `yaml:"localConfigName"`
+}
+
+func (lp *LocalProviderConfig) IsValid() error {
+	cfg := lp.ConfigManager.LocalConfigName
+	if cfg != UPGRADECONFIG_CR_NAME {
+		return fmt.Errorf("please use %s as the upgrade config name", UPGRADECONFIG_CR_NAME)
+	}
+
+	return nil
+}
--- a/pkg/localprovider/localprovider.go
+++ b/pkg/localprovider/localprovider.go
@@ -53,7 +53,7 @@ func readSpecFromConfig(ucl upgradev1alpha1.UpgradeConfigList) ([]upgradev1alpha
 	for _, u := range ucl.Items {
 		// Completed UpgradeConfigs can be ignored
 		history := u.Status.History.GetHistory(u.Spec.Desired.Version)
-		if history.Phase != upgradev1alpha1.UpgradePhaseUpgraded {
+		if history != nil && history.Phase != upgradev1alpha1.UpgradePhaseUpgraded {
 			upgradeConfigSpecs = append(upgradeConfigSpecs, u.Spec)
 		}
 	}
--- a/pkg/localprovider/localprovider.go
+++ b/pkg/localprovider/localprovider.go
@@ -1,0 +1,76 @@
+package localprovider
+
+import (
+	"context"
+	"fmt"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	upgradev1alpha1 "github.com/openshift/managed-upgrade-operator/pkg/apis/upgrade/v1alpha1"
+	"github.com/openshift/managed-upgrade-operator/util"
+)
+
+const UPGRADECONFIG_CR_NAME = "managed-upgrade-config"
+
+var log = logf.Log.WithName("upgradeconfig-localprovider")
+
+func New(c client.Client, name string) (*localProvider, error) {
+	return &localProvider{
+		client:  c,
+		cfgname: name,
+	}, nil
+}
+
+type localProvider struct {
+	client  client.Client
+	cfgname string
+}
+
+// Get checks the upgrade config on the cluster with matched name which is going to perform the upgrade
+func (l *localProvider) Get() ([]upgradev1alpha1.UpgradeConfigSpec, error) {
+	log.Info("Read the upgrade config from the cluster directly")
+
+	// Get the current UpgradeConfigs on the cluster
+	instances, err := fetchUpgradeConfigs(l.client)
+	if err != nil {
+		return nil, err
+	}
+
+	// Get Specs from the upgradeConfig
+	specs, err := readSpecFromConfig(*instances)
+	if err != nil {
+		return nil, err
+	}
+
+	return specs, nil
+}
+
+// Helper function to extract the spec from the upgradeConfig CR
+func readSpecFromConfig(ucl upgradev1alpha1.UpgradeConfigList) ([]upgradev1alpha1.UpgradeConfigSpec, error) {
+	upgradeConfigSpecs := make([]upgradev1alpha1.UpgradeConfigSpec, 0)
+
+	for _, u := range ucl.Items {
+		upgradeConfigSpecs = append(upgradeConfigSpecs, u.Spec)
+	}
+	return upgradeConfigSpecs, nil
+}
+
+// Read the CR from cluster with matched name
+func fetchUpgradeConfigs(c client.Client) (*upgradev1alpha1.UpgradeConfigList, error) {
+	instances := &upgradev1alpha1.UpgradeConfigList{}
+	ns, err := util.GetOperatorNamespace()
+	if err != nil {
+		return nil, err
+	}
+
+	err = c.List(context.TODO(), instances, client.InNamespace(ns),
+		client.MatchingFields{"metadata.name": UPGRADECONFIG_CR_NAME},
+	)
+	if err != nil {
+		log.Error(err, fmt.Sprintf("Failed to list the upgrade config with name %s", UPGRADECONFIG_CR_NAME))
+		return nil, err
+	}
+
+	return instances, nil
+}
--- a/pkg/localprovider/localprovider.go
+++ b/pkg/localprovider/localprovider.go
@@ -51,7 +51,11 @@ func readSpecFromConfig(ucl upgradev1alpha1.UpgradeConfigList) ([]upgradev1alpha
 	upgradeConfigSpecs := make([]upgradev1alpha1.UpgradeConfigSpec, 0)
 
 	for _, u := range ucl.Items {
-		upgradeConfigSpecs = append(upgradeConfigSpecs, u.Spec)
+		// Completed UpgradeConfigs can be ignored
+		history := u.Status.History.GetHistory(u.Spec.Desired.Version)
+		if history.Phase != upgradev1alpha1.UpgradePhaseUpgraded {
+			upgradeConfigSpecs = append(upgradeConfigSpecs, u.Spec)
+		}
 	}
 	return upgradeConfigSpecs, nil
 }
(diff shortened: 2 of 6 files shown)


pkg/localprovider/localprovider.go
```diff
@@ -0,0 +1,103 @@
...
+	instances, err := fetchUpgradeConfigs(l.client)
+	if err != nil {
+		return nil, err
+	}
+
+	// Remove the upgradeConfig is it has been consumed
+	for _, u := range instances.Items {
+		if isUpgradeConfigConsumed(u) {
+			log.Info(fmt.Sprintf("The upgrade for version %s has completed, remove the upgradeConfig.", u.Spec.Desired.Version))
+			err = l.client.Delete(context.TODO(), &u)
```
ravitri, line 44: I have two queries for this part. 
1. By default, MUO will reconcile against UpgradeConfig CR named `managed-upgrade-config` as such which means we do not quite care about any other UpgradeConfig. So is the iteration loop required keeping the future work for Hotfix/CVE in mind when there "might" be additional UpgradeConfig instances?
2. Should we actually delete the UpgradeConfig once the upgrade is done? Any harm in keeping it as it is? Would help for reference once upgrade is finished.

pkg/localprovider/localprovider.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// Determine if the upgradeConfig is being consumed
+func isUpgradeConfigConsumed(uc upgradev1alpha1.UpgradeConfig) bool {
+	consumed := false
+
+	desiredVersion := uc.Spec.Desired.Version
+
+	for _, h := range uc.Status.History {
+		if h.Version == desiredVersion && h.CompleteTime != nil {
```
mrbarge, line 96: Couple of things here, instead of looping over each history status, you can get the one that matches your UC's version with `uc.Status.History.GetHistory(<version>)`. See the upgradeconfig controller for a usage example.

It may also be better to check the history's `Phase` here == `UpgradePhaseUpgraded`, as that's the same logic used by the UC controller when determining if an upgrade is complete.

pkg/localprovider/localprovider.go
```diff
@@ -0,0 +1,103 @@
...
+	log.Info("Read the upgrade config from the cluster directly")
+
+	// Get the current UpgradeConfigs on the cluster
+	instances, err := fetchUpgradeConfigs(l.client)
+	if err != nil {
+		return nil, err
+	}
+
+	// Remove the upgradeConfig is it has been consumed
+	for _, u := range instances.Items {
```
mrbarge, line 41: Instead of having the provider do the delete, the `upgradeconfigmanager` can take care of this for you.

If the UpgradeConfig returned from the fetch is for a completed upgrade, just return empty slice/nil. That'll cause the upgradeconfigmanager, which call the provider, to clean up the upgradeconfig.

pkg/notifier/config.go
```diff
@@ -23,9 +24,9 @@ func (cfg *NotifierConfig) IsValid() error {
 	switch cfg.ConfigManager.Source {
 	case string(OCM):
 		return nil
+	case string(LOCAL):
+		return nil
```
ravitri, line 28: Considering this, the end user for ARO/OCP would always have to specify the source as "LOCAL" in the configmap explicitly. What about considering LOCAL as the default source if no other source (Eg. OCM) is mentioned explicitly?

pkg/specprovider/specprovider.go
```diff
@@ -46,6 +48,17 @@ func (ppb *specProviderBuilder) New(client client.Client, builder configmanager.
 			return nil, err
 		}
 		return mgr, nil
+	case "LOCAL":
```
ravitri, line 51: Considering this, the end user for ARO/OCP would always have to specify the source as "LOCAL" in the configmap explicitly. What about considering LOCAL as the default source if no other source (Eg. OCM) is mentioned explicitly?

pkg/localprovider/localprovider.go
bmeng: As Matt mentioned below, it will be deleted by upgrade-config-manager anyway, I will probably remove this part.

pkg/specprovider/specprovider.go
bmeng: We will have the configmap since there are some other places referencing the configs. 
I think it is not necessary to set a default value here.

mrbarge, line 51: For this and the other various similar comparisons, I'm starting to think we should maybe change the value evaluated by the `switch` to a `strings.ToUpper()`-ized equivalent first, so that there are no case sensitivity issues if `"local"` is used.

Probably should do it for the upgradeconfig upgrade type too, whilst I'm thinking of it.

pkg/localprovider/config.go
```diff
@@ -0,0 +1,20 @@
...
+}
+
+type ConfigManager struct {
+	LocalConfigName string `yaml:"localConfigName"`
+}
+
+func (lp *LocalProviderConfig) IsValid() error {
+	cfg := lp.ConfigManager.LocalConfigName
+	if cfg != UPGRADECONFIG_CR_NAME {
+		return fmt.Errorf("Please use %s as the upgrade config name.", UPGRADECONFIG_CR_NAME)
```
mrbarge, line 16: I definitely get why this is included - the localprovider needs to know the CR name in order to actually work, and it needs to match what the UC Manager thinks the name is. In its current form as a config item it has a few design smells, though - since the value can never deviate from what the default is, it's not a very 'configurable' item. I think the design here can be tweaked slightly to overcome that. For example, what if:

- We still retain a 'CR Name' configmap item, but name it something that's not "local" specific, ie. `upgradeConfigName` or something. 
- The *UpgradeConfigManager* should read in this 'CR Name' configmap value too, when it is initialized. If it's present, it overrides the default `UPGRADECONFIG_CR_NAME` as the name of the CR used by the UpgradeConfigManager when doing stuff. (The UC controller would also have to be updated as part of this change.)
- After doing the above - there is no reason that we need to have the check in this file to make sure that the value in config matches the default, and we can make the CR name a fully configurable item.

As it will be good to get this PR done and merged before your absence, I'm happy to leave this as-is and work on that in OSD-6918 if you like. Do let me know what you think though.

pkg/localprovider/localprovider.go
mrbarge: We can control how long the UpgradeConfigManager decides to keep an UpgradeConfig around post-upgrade via the `watchInterval` config setting in the configmap. ie. if we want it to stick around for a while, we set a watchinterval of days. So that is configurable in the current operator. 

My personal preference is that our various spec providers all behave in the same homogenous way, rather than some of them having some particular behaviour characteristics, and that there is no permanence to the CR itself, and that the UpgradeConfigManager manages that regardless of the upgrade type. I see the UpgradeConfig as analogous to something like a CertificateSigningRequest. But stressing that is just my _personal_ preference. :smile: 

So long story short, as long as the localprovider doesn't return any specs when the UC indicates that the upgrade is complete, it looks good to me.

pkg/localprovider/config.go
bmeng: I agree it should be better to make this value configurable. But I am not sure if it is in the same scope with OSD-6918? Maybe should better use another card to track?

mrbarge: Yeah, good point. It's probably better suited to another card.

bmeng: https://issues.redhat.com/browse/OSD-7073 new card created to track this

pkg/localprovider/localprovider.go
ravitri: Thanks a lot Matt !! I see your point. So the intent here is to treat UpgradeConfig as ephemeral resources relevant during upgrade time. That makes sense. However, IIUC when the provider is OCM, the necessity to delete the UC is also because of the dependency on UpgradePolicy correct? And for localprovider we do not have any such constraint. My point of view was that if ARO or OCP end user explicitly creates UC and doesn't find it after upgrade if we delete it, we can log it in MUO and also provide documented steps to know upgrade status. Else not delete the UC at all if it's ok to keep it.

pkg/specprovider/config.go
```diff
@@ -2,10 +2,12 @@ package specprovider
 
 import (
 	"fmt"
+	"strings"
 )
 
 const (
-	OCM ConfigManagerSource = "OCM"
+	OCM   ConfigManagerSource = "OCM"
```
dofinn, line 9: can we manage these in one place? 
https://github.com/openshift/managed-upgrade-operator/pull/225/files#diff-fb43714e995d746a39a260970fd063b13f96e634e146910ea6d3e2630c7057e5R6-R7

pkg/localprovider/localprovider.go
```diff
@@ -0,0 +1,76 @@
...
+
+	return specs, nil
+}
+
+// Helper function to extract the spec from the upgradeConfig CR
+func readSpecFromConfig(ucl upgradev1alpha1.UpgradeConfigList) ([]upgradev1alpha1.UpgradeConfigSpec, error) {
+	upgradeConfigSpecs := make([]upgradev1alpha1.UpgradeConfigSpec, 0)
+
+	for _, u := range ucl.Items {
+		upgradeConfigSpecs = append(upgradeConfigSpecs, u.Spec)
```
mrbarge, line 54: Could we change this so that it only returns specs that aren't upgraded? That will cause the UpgradeConfigManager to remove the upgraded UpgradeConfig on its next watch cycle.

```suggestion
		// Completed UpgradeConfigs can be ignored
		history := u.Status.History.GetHistory(u.Spec.Desired.Version)
		if history.Phase != upgradev1alpha1.UpgradePhaseUpgraded {
			upgradeConfigSpecs = append(upgradeConfigSpecs, u.Spec)
		}
```

(I gave the PR a test with this proposed change)

Other than that, everything looks great and I'll be happy to lgtm it.

pkg/localprovider/localprovider.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// Helper function to extract the spec from the upgradeConfig CR
+func readSpecFromConfig(ucl upgradev1alpha1.UpgradeConfigList) ([]upgradev1alpha1.UpgradeConfigSpec, error) {
+	upgradeConfigSpecs := make([]upgradev1alpha1.UpgradeConfigSpec, 0)
+
+	for _, u := range ucl.Items {
+		// Completed UpgradeConfigs can be ignored
+		history := u.Status.History.GetHistory(u.Spec.Desired.Version)
+		if history.Phase != upgradev1alpha1.UpgradePhaseUpgraded {
```
mrbarge, line 56: I just thought of one more thing that I forgot, sorry. :joy: 
To be panic safe it should maybe check that the history returned isn't nil first.

```suggestion
		if history != nil && history.Phase != upgradev1alpha1.UpgradePhaseUpgraded {
```

pkg/localprovider/localprovider.go
mrbarge: We can certainly keep the UC around for a very long time post-upgrade with configuration of the `watchInterval` in the configmap (granted, it would still get cleaned up on operator restart).

The clean-up of the UC after upgrade completion on the OSD side isn't really tied to OCM at all or necessary to make that functionality work, it's just the design path we went down for how the operator behaves and treats the resource. (nothing would break if the UC stuck around in OSD land instead of being cleaned up)

We should revisit this closer to the launch date on ARO if it makes sense for how ARO SRE want to interact with the operator, and it will be a simple thing to change - in the meantime I'm keen to keep behaviour as consistent as possible from a maintainability perspective.

mrbarge:
/lgtm
Let's roll! :tada:

PR #227, opened 2021-05-06
[OSD-7104] Change scheduling of MUO on master nodes

ravitri:
### What type of PR is this?
_Deployment change_

As per [OSD-7104](https://issues.redhat.com/browse/OSD-7104), this PR would change the scheduling of MUO pod from infra node to master node (similar to CVO). The reason for this change is because upgrades got stuck when MUO pod is OOM killed on infra node when Prometheus pod consumes high memory. Since MUO pod currently does not have any priorityClass set, it has a low priority which easily qualifies for it to be OOM killed and possibly compete with customer workloads due to "preferred" affinity for infra node. This PR would separate out MUO scheduling to master nodes which doesn't interfere with infra/customer workloads at all. Considering the resource consumption is low for MUO pod, this was considered acceptable.

cc: @cblecker , @jewzaam 

_Fixes #[OSD-7104](https://issues.redhat.com/browse/OSD-7104)_

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -16,16 +16,18 @@ spec:
       serviceAccountName: managed-upgrade-operator
       affinity:
         nodeAffinity:
-          preferredDuringSchedulingIgnoredDuringExecution:
-          - preference:
-              matchExpressions:
-              - key: node-role.kubernetes.io/infra
-                operator: Exists
-            weight: 1
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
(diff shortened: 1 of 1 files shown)


ravitri:
/hold 
To unhold after discussion in https://issues.redhat.com/browse/OSD-7104 is acked and cleared by KB.

ravitri:
/unhold
Change has been reviewed, approved and tested so removing the hold

mrbarge:
I've tested this in staging and it's working well. Let's roll!
/lgtm

PR #230, opened 2021-05-13
[OSD-7106] make scale worker nodes configurable

bmeng:
### What type of PR is this?
_feature_

make the scale up/down extra worker nodes configurable since it is not needed in some conditions, like ARO

_Fixes #[OSD-7106](https://issues.redhat.com/browse/OSD-7106) #OSD-7107_

### DIFF
--- a/deploy/crds/upgrade.managed.openshift.io_upgradeconfigs_crd.yaml
+++ b/deploy/crds/upgrade.managed.openshift.io_upgradeconfigs_crd.yaml
@@ -53,7 +53,7 @@ spec:
               format: int32
               type: integer
             capacityReservation:
-              description: Specify if scale up extra node for capacity reservation before upgrade starts is needed
+              description: Specify if scaling up an extra node for capacity reservation before upgrade starts is needed
               type: boolean
             desired:
               description: Specify the desired OpenShift release
@@ -99,7 +99,6 @@ spec:
               type: string
           required:
             - PDBForceDrainTimeout
-            - capacityReservation
             - desired
             - type
             - upgradeAt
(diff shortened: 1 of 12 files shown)


deploy/crds/upgrade.managed.openshift.io_upgradeconfigs_crd.yaml
```diff
@@ -96,6 +99,7 @@ spec:
               type: string
           required:
             - PDBForceDrainTimeout
+            - capacityReservation
```
mrbarge, line 102: Given this will be a mandatory field, have you / can we test what will happen if a cluster has an existing `UpgradeConfig` CR without this field in it, when OLM updates the operator/CRD on a cluster to this version? (before the operator/upgrade config manager even starts up).

deploy/crds/upgrade.managed.openshift.io_upgradeconfigs_crd.yaml
```diff
@@ -52,6 +52,9 @@ spec:
               description: The maximum grace period granted to a node whose drain is blocked by a Pod Disruption Budget, before that drain is forced. Measured in minutes.
               format: int32
               type: integer
+            capacityReservation:
+              description: Specify if scale up extra node for capacity reservation before upgrade starts is needed
```
mrbarge, line 56: Minor nit
```suggestion
              description: Specify if scaling up an extra node for capacity reservation before upgrade starts is needed
```

deploy/crds/upgrade.managed.openshift.io_upgradeconfigs_crd.yaml
bmeng: Hmm, it will get error like this immediately when the CRD updated with new value
```
"error":"UpgradeConfig.upgrade.managed.openshift.io \"managed-upgrade-config\" is invalid: spec.capacityReservation: Required value"```

PR #233, opened 2021-05-21
Boilerplate: Update to 5fea264b458ec33913afb04eda6ff87d8013d9f4

clcollins:
Conventions:
- openshift/golang-osd-operator: Update
---
https://github.com/openshift/boilerplate/compare/0ba6566d544d0df9993a92b2286c131eb61f3e88...5fea264b458ec33913afb04eda6ff87d8013d9f4

commit: e6753c2a4252695328187dad83370e63f308e355
author: Christopher Collins
Updates the version of the openshift-operator-registry image

Updates the version of the openshift-operator-registry image to 4.8.0,
to address several high vulnerabilities found in the 4.7.0 image.

REF: [OSD-6831](https://issues.redhat.com/browse/OSD-6831)

Signed-off-by: Christopher Collins <collins.christopher@gmail.com>

commit: c35da9dd532d5ed389e4e93f83c75fe9e51e98c9
author: Eric Fried
Fix generated-files-checker for go 1.16

go 1.16 is stricter about go.sum being up to date for things like `go
-list`. Fix up the test (case & project) accordingly.

Also add test loops for missing supported asdk versions. Not sure when
we missed those.

Also fix the backing image -- it was hacking the go lib/cache
permissions too early.

Also make it possible to override the backing image used by
`container-make`, to ease debugging when writing commits like this.

commit: 0acfc9d8d31942977df26f49ec043038d2fd3320
author: Rob Rati
Update to use golang 1.16

commit: c49cf6f755ffab5c6e2ce6e4118b743e83093248
author: Eric Fried
Fix unbound local variable error

...noted in https://ci.int.devshift.net/job/app-sre-deployment-validation-operator-gh-build-master/60/console

commit: 3bcb750bd6dc9120f5458aa44d82fff6973d2171
author: Eric Fried
Improve `subscriber report release`

This commit improves the `subscriber report release` tool in the
following ways:
- It now requires a list of subscribers, or `ALL`, with the same
semantics as `subscriber propose update`.
- It now produces a diff with which you can `patch -R` what's in the
release repository to make it conform.

commit: 9062716c4df7b01b72ba4308e106b4c932ab5649
author: Christopher Collins
Adds .gitattributes to unacceptable_deltas "ignores" list

With a recent update to boilerplate, a .gitattributes file is created
during the bootstrap process.  This is causing failures in the
"framework/04-update-from-master-and-revert" test, which currently
doesn't expect the .gitattributes file to exist, uncommitted, after the
test bootstrap.

This adds the .gitattributes file to the "ignores" list in the
unacceptable_deltas function of boilerplate/update.

Also runs `./boilerplate/_lib/boilerplate-commit` to "clean" before the
next test run.

Signed-off-by: Christopher Collins <collins.christopher@gmail.com>

commit: 69519b47fcbd219aa4d30aeb28377c2113ea8272
author: Eric Fried
Secure freeze-check, enable hiding boilerplate deltas

- Document the "trust and ignore" philosophy that subscribers should
adopt when reviewing PRs that include boilerplate updates.
- Secure `freeze-check` so you actually _can_ trust such changes.
- Lay down .gitattributes content such that, if a subscriber decides to
suppress boilerplate deltas by default, certain files will still always
be shown:
  - Because an attacker could attempt to subvert `freeze-check` by
  modifying that script or its dependencies, we explicitly exclude those
  from the hiding-by-default.
  - Because an attacker could try to subvert the hiding by appending
  entries to .gitattributes, we exclude THAT file from the
  hiding-by-default as well.

Security is hard.


mrbarge:
The issue I raised turned out to be unrelated to this PR. And I've done a couple of successful test runs. Think we're good to let this land.
/lgtm
/hold cancel

PR #236, opened 2021-05-28
OSD-7325 remove upgrade verification steps

bmeng:
### What type of PR is this?
_feature/cleanup_

Remove the post verification step
Currently we use the post verification to check the replica count for DS and RS in the particular namespaces, and we are going to fix those if the replica counts mismatch during the upgrade. Which is the same we deal with the normal cluster issue.
To make the upgrade process itself more stable, we should let the cluster monitoring to catch those ds/rs count mismatch. And fix them outside the cluster upgrade loop.

_Fixes [OSD-7325](https://issues.redhat.com/browse/OSD-7325)_


mrbarge:
Haven't tested it, but it looks good. I'll give others a chance to review also before a merge.

PR #241, opened 2021-06-07
OSD-7095 - Adding localprovider tests

ravitri:
### What type of PR is this?
_(test)_

This PR is to add the tests for `localprovider` package for MUO.

_Fixes #[OSD-7095](https://issues.redhat.com/browse/OSD-7095)_

None

- [x] Ran `make test` command locally to validate code changes

### DIFF
--- a/pkg/localprovider/localprovider_suite_test.go
+++ b/pkg/localprovider/localprovider_suite_test.go
@@ -1,4 +1,4 @@
-package localprovider_test
+package localprovider
 
 import (
 	"testing"
--- a/pkg/localprovider/localprovider_suite_test.go
+++ b/pkg/localprovider/localprovider_suite_test.go
@@ -1,0 +1,13 @@
+package localprovider_test
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestLocalprovider(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Localprovider Suite")
+}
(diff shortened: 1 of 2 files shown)


pkg/localprovider/localprovider_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package localprovider_test
```
bmeng, line 1: package should be localprovider

pkg/localprovider/localprovider_test.go
```diff
@@ -0,0 +1,158 @@
...
+			cfgname: TEST_LOCAL_UPGRADECONFIG_NAME,
+		}
+
+		upgradeConfigList = v1alpha1.UpgradeConfigList{
+			Items: []v1alpha1.UpgradeConfig{
+				{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "uc1",
+					},
+					Spec: v1alpha1.UpgradeConfigSpec{
```
bmeng, line 52: Can we just use the defined spec contains, like:
```
Spec: upgradeConfigSpec,
```
If they have the same values.

pkg/localprovider/localprovider_suite_test.go
ravitri: Thanks Meng Bo! Done.

pkg/localprovider/localprovider_test.go
ravitri: Nice suggestion. Would be cleaner. Done.

mrbarge:
One of the behavioural characteristics of the localprovider is that it doesn't return a Upgrade Config spec if the UpgradeConfig's history indicates that it is upgraded. There is the opportunity for having a test around this that is presently lacking.

ravitri:
Thanks a lot Matt! Have added the required test.

PR #243, opened 2021-06-07
localprovider doc update

a7vicky:

### DIFF
--- a/docs/configmanager.md
+++ b/docs/configmanager.md
@@ -10,7 +10,7 @@ Currently, the following sources are supported:
 | Source | Description |
 | --- | --- |
 | `OCM` | Retrieve an UpgradeConfig from the OpenShift Cluster Manager [`upgrade_policies`](https://api.openshift.com/#/default/get_api_clusters_mgmt_v1_clusters__cluster_id__upgrade_policies) API |
-| `LOCAL` | Using UpgradeConfig CR localy on the OpenShift Cluster|
+| `LOCAL` | Using UpgradeConfig CR locally on the OpenShift Cluster|
 
 ## Configuring an UpgradeConfig Manager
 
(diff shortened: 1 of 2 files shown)


docs/configmanager.md
```diff
@@ -10,6 +10,7 @@ Currently, the following sources are supported:
 | Source | Description |
 | --- | --- |
 | `OCM` | Retrieve an UpgradeConfig from the OpenShift Cluster Manager [`upgrade_policies`](https://api.openshift.com/#/default/get_api_clusters_mgmt_v1_clusters__cluster_id__upgrade_policies) API |
+| `LOCAL` | Using UpgradeConfig CR localy on the OpenShift Cluster|
```
dofinn, line 13:
```suggestion
| `LOCAL` | Using UpgradeConfig CR locally on the OpenShift Cluster|
```

docs/configmanager.md
```diff
@@ -38,3 +39,13 @@ configManager:
...
 ```
+
+### LOCAL UpgradeConfig Manager
+
+The following configuration fields must be set:
+
+| Field | Description | Example |
+| --- | --- | --- |
+| `source` | Indicates the type of config manager being used | `LOCAL` |
+| `LocalConfigName` | Name of the Local config being used | `example-upgrade-config` |
```
dofinn, line 50: @mrbarge was the config a singleton? "managed-upgrade-config" ?

a7vicky: @dofinn Thanks for review. Yes it seems [upgradeconfig CR name](https://github.com/openshift/managed-upgrade-operator/blob/3518984ff470946edfb8156f1ac388c4c0be98ec/pkg/localprovider/localprovider.go#L14) (managed-upgrade-config) is fixed which is looked by MUO.

https://github.com/openshift/managed-upgrade-operator/blob/master/pkg/localprovider/localprovider.go#L63-L80

docs/configmanager.md
a7vicky: @dofinn  I have fixed the typo.

PR #244, opened 2021-06-08
MUO [metrics, FAQ] doc update 

a7vicky:

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -14,7 +14,7 @@ The `UpgradeConfig` can be created directly on the cluster which is not managed 
 
 For OpenShift Dedicated cluster you should not create `UpgradeConfig` directly. The process for scheduling an upgrade is automated and achieved using the [upgrade_policies API](https://api.openshift.com/#/default/get_api_clusters_mgmt_v1_clusters__cluster_id__upgrade_policies). 
 
-The process for SRE to manage the creation and distribution of `UpgradeConfig` custom resources is documented in [SOPs](https://github.com/openshift/ops-sop/blob/master/v4/howto/managed-upgrade.md).
+The process for creating `upgrade-policy` is documented in [SOPs](https://github.com/openshift/ops-sop/blob/master/v4/howto/managed-upgrade.md).
 
 ## Custom Resource Definitions
 
--- a/docs/design.md
+++ b/docs/design.md
@@ -12,7 +12,7 @@ The operator is primarily driven through an `UpgradeConfig` custom resource, whi
 
 The `UpgradeConfig` can be created directly on the cluster which is not managed via Hive. Also this method can be used for development/testing purposes
 
-For OpenShift Dedicated deployments, the `UpgradeConfig` is distributed via a [Hive SelectorSyncSet](https://github.com/openshift/hive/blob/master/docs/syncset.md) and managed by OpenShift SRE.
+For OpenShift Dedicated deployments, the `UpgradeConfig` is distributed via a [Hive SyncSet](https://github.com/openshift/hive/blob/master/docs/syncset.md) and managed by OpenShift SRE.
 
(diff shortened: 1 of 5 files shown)


docs/design.md
```diff
@@ -12,7 +12,7 @@ The operator is primarily driven through an `UpgradeConfig` custom resource, whi
 
 The `UpgradeConfig` can be created directly on the cluster which is not managed via Hive. Also this method can be used for development/testing purposes
 
-For OpenShift Dedicated deployments, the `UpgradeConfig` is distributed via a [Hive SelectorSyncSet](https://github.com/openshift/hive/blob/master/docs/syncset.md) and managed by OpenShift SRE.
+For OpenShift Dedicated deployments, the `UpgradeConfig` is distributed via a [Hive SyncSet](https://github.com/openshift/hive/blob/master/docs/syncset.md) and managed by OpenShift SRE.
```
mrbarge, line 15: We should get rid of this line entirely now - this hasn't been the case since late last year.

docs/metrics.md
```diff
@@ -11,3 +11,13 @@ The Managed Upgrade Operator reports the following metrics:
 - `upgrade_worker_start`: The start time of the worker upgrades
 - `upgrade_worker_completion`: The completion time of the worker upgrades
 - `upgrade_complete`: The completion time of the managed upgrade
+
+## Metrics use to alert SRE
```
mrbarge, line 15: This section _may_ end up changing in the very near future when OSD-6778 is finished. But it's fine for now.

docs/design.md
a7vicky: Thanks for review @mrbarge.
I have updated the doc and removed the entire line from the doc.

docs/faq.md
```diff
@@ -14,11 +14,11 @@ Currently this is a manual process. We are working on dashboards and other metri
 
 **Does MUO reserve compute capacity?**
 
-Yes, if `capacityReservation: true` is enabled in the upgradeconfig CR. MUO will create a temporary +1 to every worker `machinesets` within the cluster. In multi availability zones, this is true for each zone.
+Yes, if `capacityReservation` is not disabled in the upgradeconfig CR. MUO will create a temporary +1 to every worker `machinesets` within the cluster. In multi availability zones, this is true for each zone.
```
mrbarge, line 17: If it is not specified at all when using a `local` provider, is it still disabled?

This might need a little rewording also because it reads a little bit inaccurately. MUO does not modify the count of existing worker machinesets. It creates a new `-upgrade` worker machineset for each availability zone with a size of 1.

T0MASD, line 17:
```suggestion
Yes, if `capacityReservation` in the upgradeconfig CR is set to `true`. MUO will create a temporary +1 to every worker `machinesets` within the cluster. In multi availability zones, this is true for each zone.
```

docs/faq.md
```diff
@@ -14,11 +14,11 @@ Currently this is a manual process. We are working on dashboards and other metri
...
 
 **Does MUO maintain correct instance types for each machine pool?**
 
-Yes, if `capacityReservation: true` is enabled in the upgradeconfig CR. MUO creates the extra compute based on the found instance types of the current `machinesets`.
+Yes, if `capacityReservation` is not disabled in the upgradeconfig CR. MUO creates the extra compute based on the found instance types of the current `machinesets`.
```
T0MASD, line 21:
```suggestion
Yes, if `capacityReservation` in the upgradeconfig CR is set to `true`. MUO creates the extra compute based on the found instance types of the current `machinesets`.
```

T0MASD (commented):
Please review my suggestions and get another ack

docs/faq.md
a7vicky: Thank you @mrbarge for review. I have updated the details for local provider in case `spec.capacityReservation` is missing in the upgradeconfig CR. Also  I have rephrased wording for `upgrade` worker machinesets.

docs/faq.md
a7vicky: Thanks @T0MASD for review. I have updated the doc based on your suggestion.

mrbarge:
Thanks for this!
/lgtm

PR #245, opened 2021-06-09
make the namespace in health check configurable

bmeng:
### What type of PR is this?
_feature_

Make the ignored namespace configurable in health check, so that we could change it easily 

_Fixes OSD-5943_

### DIFF
--- a/pkg/upgraders/osd/config.go
+++ b/pkg/upgraders/osd/config.go
@@ -59,7 +59,8 @@ type scaleConfig struct {
 }
 
 type healthCheck struct {
-	IgnoredCriticals []string `yaml:"ignoredCriticals"`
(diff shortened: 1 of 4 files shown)


mrbarge:
Seems fine to me (haven't tested). 
Now that it's configurable I'm on the fence about whether we should be setting any 'defaults' at all if it's not configured. If we end up making 'steps' generic across upgraders that point becomes more of an issue. Not so much right now though.
So in general 👍 from a review standpoint.

T0MASD (commented):
Thanks, looks good!

bmeng:
adding the defaults here, just to make sure that without the variables setting in the configmap, the MUO should still behavior as usual.
if we do not have the default, we should make the change in MCC prior to this one.

PR #253, opened 2021-06-28
Add support for upgrades using image digest reference

ravitri:
### What type of PR is this?
_(feature)_

With regard to Jira card [OSD-7000](https://issues.redhat.com/browse/OSD-7000), this PR serves the requirement to initiate an upgrade using an image digest reference as well. This would help users to initiate an upgrade when the cluster has no upstream URL or channel configured in `ClusterVersion`. 

_Fixes #_  [OSD-7000](https://issues.redhat.com/browse/OSD-7000)

- Currently, the scope of using an image digest is tied up with a mandatory `version` field as well which SHOULD match the version that's in the image's release metadata. This manual verification should be done by the user.
- In future, this verification will be automated and the `image` will be referenced without associated `version`. 
- Will add tests for validation as a separate PR.

- [x] Tested two upgrades using `image` and `channel`.

### DIFF
--- a/deploy/crds/upgrade.managed.openshift.io_upgradeconfigs_crd.yaml
+++ b/deploy/crds/upgrade.managed.openshift.io_upgradeconfigs_crd.yaml
@@ -61,12 +61,12 @@ spec:
                 channel:
                   description: Channel used for upgrades
                   type: string
+                image:
+                  description: Image reference used for upgrades
+                  type: string
                 version:
                   description: Version of openshift release
                   type: string
-              required:
-                - channel
-                - version
               type: object
             type:
               description: Type indicates the ClusterUpgrader implementation to use to perform an upgrade of the cluster
--- a/pkg/clusterversion/cv.go
+++ b/pkg/clusterversion/cv.go
@@ -3,7 +3,9 @@ package clusterversion
 import (
 	"context"
 	"fmt"
+	"strings"
 
+	"github.com/go-logr/logr"
 	configv1 "github.com/openshift/api/config/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -15,6 +17,7 @@ import (
 var (
 	// OSD_CV_NAME is the name of cluster version singleton
 	OSD_CV_NAME = "version"
+	logger      logr.Logger
 )
 
 // ClusterVersion interface enables implementations of the ClusterVersion
@@ -74,8 +77,9 @@ func (c *clusterVersionClient) EnsureDesiredConfig(uc *upgradev1alpha1.UpgradeCo
 
 	// Commence upgrade using version and channel
 	// Move the cluster to the same channel first
-	if len(desired.Channel) > 0 && len(desired.Version) > 0 {
+	if !empty(desired.Channel) && !empty(desired.Version) {
 		if clusterVersion.Spec.Channel != desired.Channel {
+			logger.Info(fmt.Sprintf("Setting ClusterVersion to Channel %s Version %s", desired.Channel, desired.Version))
 			clusterVersion.Spec.Channel = desired.Channel
 			err = c.client.Update(context.TODO(), clusterVersion)
 			if err != nil {
@@ -109,14 +113,21 @@ func (c *clusterVersionClient) EnsureDesiredConfig(uc *upgradev1alpha1.UpgradeCo
 	}
 
 	// Commence upgrade using Image
-	if len(desired.Image) > 0 {
-		if clusterVersion.Spec.DesiredUpdate.Image != desired.Image {
-			clusterVersion.Spec.DesiredUpdate.Image = desired.Image
+	// TODO: Remove dependency on version as per OSD-7609
+	if !empty(desired.Image) && !empty(desired.Version) && empty(desired.Channel) {
+		if clusterVersion.Spec.DesiredUpdate == nil || (clusterVersion.Spec.DesiredUpdate.Image != desired.Image) {
+			logger.Info(fmt.Sprintf("Setting ClusterVersion to Image %s", desired.Image))
+			clusterVersion.Spec.DesiredUpdate = &configv1.Update{Image: desired.Image}
 			err = c.client.Update(context.TODO(), clusterVersion)
 			if err != nil {
 				return false, err
 			}
 		}
+	}
+
+	// If neither (version+channel) nor (image) is defined, throw error.
+	if (empty(desired.Version) && empty(desired.Channel)) && empty(desired.Image) {
+		return false, fmt.Errorf("need atleast (version+channel) or (image) defined in UpgradeConfig")
 	}
 
 	return true, nil
@@ -166,25 +177,15 @@ func (c *clusterVersionClient) HasUpgradeCompleted(cv *configv1.ClusterVersion, 
 
 // isEqualVersion compare the upgrade version state for cv and uc
 func isEqualVersion(cv *configv1.ClusterVersion, uc *upgradev1alpha1.UpgradeConfig) bool {
-	if cv.Spec.DesiredUpdate != nil &&
-		cv.Spec.DesiredUpdate.Version == uc.Spec.Desired.Version {
-		return true
-	}
-
-	return false
+	return cv.Spec.DesiredUpdate != nil && (cv.Spec.DesiredUpdate.Version == uc.Spec.Desired.Version)
 }
 
 // isEqualImage compare the upgrade version state for cv and uc
 func isEqualImage(cv *configv1.ClusterVersion, uc *upgradev1alpha1.UpgradeConfig) bool {
-	if cv.Spec.DesiredUpdate != nil &&
-		cv.Spec.DesiredUpdate.Image == uc.Spec.Desired.Image {
-		return true
-	}
-
-	return false
+	return cv.Spec.DesiredUpdate != nil && (cv.Spec.DesiredUpdate.Image == uc.Spec.Desired.Image)
 }
 
-// hasUpgradeCommenced checks if the upgrade has commenced based on version or image
+// HasUpgradeCommenced checks if the upgrade has commenced based on version or image
 func (c *clusterVersionClient) HasUpgradeCommenced(uc *upgradev1alpha1.UpgradeConfig) (bool, error) {
 
 	clusterVersion, err := c.GetClusterVersion()
@@ -192,15 +193,20 @@ func (c *clusterVersionClient) HasUpgradeCommenced(uc *upgradev1alpha1.UpgradeCo
 		return false, err
 	}
 
-	if len(uc.Spec.Desired.Version) > 0 && len(uc.Spec.Desired.Channel) > 0 {
+	if !empty(uc.Spec.Desired.Version) && !empty(uc.Spec.Desired.Channel) && empty(uc.Spec.Desired.Image) {
 		if !isEqualVersion(clusterVersion, uc) {
 			return false, nil
+		} else {
+			logger.Info(fmt.Sprintf("ClusterVersion is already set to Channel %s Version %s", uc.Spec.Desired.Channel, uc.Spec.Desired.Version))
 		}
 	}
 
-	if len(uc.Spec.Desired.Version) > 0 && len(uc.Spec.Desired.Image) > 0 {
+	// TODO: Remove version dependency after work on OSD-7609
+	if !empty(uc.Spec.Desired.Version) && !empty(uc.Spec.Desired.Image) && empty(uc.Spec.Desired.Channel) {
 		if !isEqualImage(clusterVersion, uc) {
 			return false, nil
+		} else {
+			logger.Info(fmt.Sprintf("ClusterVersion is already set to Image %s", uc.Spec.Desired.Image))
 		}
 	}
 
@@ -236,4 +242,9 @@ func GetCurrentVersion(clusterVersion *configv1.ClusterVersion) (string, error) 
 	}
 
 	return gotVersion, nil
+}
+
+// empty function checks if a given string is empty or not.
+func empty(s string) bool {
+	return strings.TrimSpace(s) == ""
 }
--- a/pkg/clusterversion/cv.go
+++ b/pkg/clusterversion/cv.go
@@ -10,14 +10,15 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
 	upgradev1alpha1 "github.com/openshift/managed-upgrade-operator/pkg/apis/upgrade/v1alpha1"
 )
 
 var (
 	// OSD_CV_NAME is the name of cluster version singleton
-	OSD_CV_NAME = "version"
-	logger      logr.Logger
+	OSD_CV_NAME             = "version"
+	logger      logr.Logger = logf.Log.WithName("clusterversion")
 )
 
 // ClusterVersion interface enables implementations of the ClusterVersion
--- a/pkg/clusterversion/cv.go
+++ b/pkg/clusterversion/cv.go
@@ -22,7 +22,7 @@ var (
 type ClusterVersion interface {
 	GetClusterVersion() (*configv1.ClusterVersion, error)
 	HasUpgradeCommenced(*upgradev1alpha1.UpgradeConfig) (bool, error)
-	EnsureDesiredVersion(uc *upgradev1alpha1.UpgradeConfig) (bool, error)
+	EnsureDesiredConfig(uc *upgradev1alpha1.UpgradeConfig) (bool, error)
 	HasUpgradeCompleted(*configv1.ClusterVersion, *upgradev1alpha1.UpgradeConfig) bool
 	HasDegradedOperators() (*HasDegradedOperatorsResult, error)
 }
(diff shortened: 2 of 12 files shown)


pkg/upgraders/osd/upgrader.go
```diff
@@ -126,9 +126,14 @@ func PreClusterHealthCheck(c client.Client, cfg *osdUpgradeConfig, scaler scaler
...
+
 	desired := upgradeConfig.Spec.Desired
 	if upgradeCommenced {
-		logger.Info(fmt.Sprintf("ClusterVersion is already set to Channel %s Version %s, skipping %s", desired.Channel, desired.Version, upgradev1alpha1.UpgradePreHealthCheck))
+		if len(desired.Channel) > 0 && len(desired.Version) > 0 {
+			logger.Info(fmt.Sprintf("ClusterVersion is already set to Channel %s Version %s, skipping %s", desired.Channel, desired.Version, upgradev1alpha1.UpgradePreHealthCheck))
+		} else if len(desired.Image) > 0 && len(desired.Version) > 0 {
+			logger.Info(fmt.Sprintf("ClusterVersion is already set to Image %s, skipping %s", desired.Image, upgradev1alpha1.UpgradePreHealthCheck))
+		}
 		return true, nil
```
bmeng, line 137: we should either fail here if the version is not provided or make it required in the type definition

pkg/upgraders/osd/upgrader.go
```diff
@@ -126,9 +126,14 @@ func PreClusterHealthCheck(c client.Client, cfg *osdUpgradeConfig, scaler scaler
 	if err != nil {
 		return false, err
 	}
+
 	desired := upgradeConfig.Spec.Desired
 	if upgradeCommenced {
-		logger.Info(fmt.Sprintf("ClusterVersion is already set to Channel %s Version %s, skipping %s", desired.Channel, desired.Version, upgradev1alpha1.UpgradePreHealthCheck))
+		if len(desired.Channel) > 0 && len(desired.Version) > 0 {
```
bmeng, line 132: can we just reuse this in some way instead of copy the same block multiple times?

deploy/crds/upgrade.managed.openshift.io_upgradeconfigs_crd.yaml
```diff
@@ -61,12 +61,12 @@ spec:
                 channel:
                   description: Channel used for upgrades
                   type: string
+                image:
+                  description: Image reference used for upgrades
```
dofinn, line 65: @ravitri have you tested this API change with OLM? https://github.com/openshift/managed-upgrade-operator/blob/master/docs/catalogsource.md

pkg/clusterversion/cv.go
```diff
@@ -64,45 +64,60 @@ func (c *clusterVersionClient) GetClusterVersion() (*configv1.ClusterVersion, er
...
 	desired := uc.Spec.Desired
-	if clusterVersion.Spec.Channel != desired.Channel {
-		clusterVersion.Spec.Channel = desired.Channel
-		err = c.client.Update(context.TODO(), clusterVersion)
-		if err != nil {
-			return false, err
+
+	// Commence upgrade using version and channel
+	// Move the cluster to the same channel first
+	if len(desired.Channel) > 0 && len(desired.Version) > 0 {
```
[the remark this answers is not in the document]
dofinn: These attributes are strings, not arrays. Why not validate as `!= ""` ?

pkg/clusterversion/cv.go
```diff
@@ -64,45 +64,60 @@ func (c *clusterVersionClient) GetClusterVersion() (*configv1.ClusterVersion, er
...
 		}
 	}
 
-	// The CVO may need time sync the version before launching the upgrade
-	updateAvailable := false
-	for _, update := range clusterVersion.Status.AvailableUpdates {
-		if update.Version == desired.Version && update.Image != "" {
-			updateAvailable = true
+	// Commence upgrade using Image
+	if len(desired.Image) > 0 {
```
dofinn, line 112: we should be more defensive here and validate that `desired.Channel` is an empty string and desired.Version is not an empty string?

pkg/clusterversion/cv.go
```diff
@@ -159,16 +174,34 @@ func isEqualVersion(cv *configv1.ClusterVersion, uc *upgradev1alpha1.UpgradeConf
...
 
-// hasUpgradeCommenced checks if the upgrade has commenced
+// isEqualImage compare the upgrade version state for cv and uc
+func isEqualImage(cv *configv1.ClusterVersion, uc *upgradev1alpha1.UpgradeConfig) bool {
+	if cv.Spec.DesiredUpdate != nil &&
+		cv.Spec.DesiredUpdate.Image == uc.Spec.Desired.Image {
+		return true
+	}
+
+	return false
```
dofinn, line 184:
```suggestion
	return cv.Spec.DesiredUpdate != nil && cv.Spec.DesiredUpdate.Image == uc.Spec.Desired.Image 
```

pkg/clusterversion/cv.go
```diff
@@ -159,16 +174,34 @@ func isEqualVersion(cv *configv1.ClusterVersion, uc *upgradev1alpha1.UpgradeConf
...
 func (c *clusterVersionClient) HasUpgradeCommenced(uc *upgradev1alpha1.UpgradeConfig) (bool, error) {
 
 	clusterVersion, err := c.GetClusterVersion()
 	if err != nil {
 		return false, err
 	}
 
-	if !isEqualVersion(clusterVersion, uc) {
-		return false, nil
+	if len(uc.Spec.Desired.Version) > 0 && len(uc.Spec.Desired.Channel) > 0 {
```
dofinn, line 195: As above , would it be more appropriate to check these attributes with `!= ""` ?

pkg/clusterversion/cv.go
```diff
@@ -159,16 +174,34 @@ func isEqualVersion(cv *configv1.ClusterVersion, uc *upgradev1alpha1.UpgradeConf
...
+		if !isEqualVersion(clusterVersion, uc) {
+			return false, nil
+		}
+	}
+
+	if len(uc.Spec.Desired.Version) > 0 && len(uc.Spec.Desired.Image) > 0 {
```
dofinn, line 201: As above , would it be more appropriate to check these attributes with `!= "" ?`

dofinn (commented):
Great job @ravitri  :partying_face:

pkg/clusterversion/cv.go
```diff
@@ -64,45 +64,60 @@ func (c *clusterVersionClient) GetClusterVersion() (*configv1.ClusterVersion, er
...
 
-	// The CVO may need time sync the version before launching the upgrade
-	updateAvailable := false
-	for _, update := range clusterVersion.Status.AvailableUpdates {
-		if update.Version == desired.Version && update.Image != "" {
-			updateAvailable = true
+	// Commence upgrade using Image
+	if len(desired.Image) > 0 {
+		if clusterVersion.Spec.DesiredUpdate.Image != desired.Image {
+			clusterVersion.Spec.DesiredUpdate.Image = desired.Image
```
mrbarge, line 114: This needs to check that `clusterVersion.Spec.DesiredUpdate` is not nil (which will be the case on a fresh cluster), or else it will panic.
And it needs to initialize `DesiredUpdate` if it is nil, or else it will panic when setting `Image`.

Something like:
```suggestion
		if clusterVersion.Spec.DesiredUpdate == nil || (clusterVersion.Spec.DesiredUpdate.Image != desired.Image) {
			clusterVersion.Spec.DesiredUpdate = &configv1.Update{Image: desired.Image}
```

pkg/upgraders/osd/upgrader.go
ravitri: Done.. Moved the logging part to clusterversion package then the upgrader..

pkg/clusterversion/cv.go
```diff
@@ -64,45 +64,60 @@ func (c *clusterVersionClient) GetClusterVersion() (*configv1.ClusterVersion, er
...
-	if !updateAvailable {
-		return false, nil
-	}
-
-	clusterVersion.Spec.Overrides = []configv1.ComponentOverride{}
-	clusterVersion.Spec.DesiredUpdate = &configv1.Update{Version: uc.Spec.Desired.Version}
-	err = c.client.Update(context.TODO(), clusterVersion)
-	if err != nil {
-		return false, err
-	}
```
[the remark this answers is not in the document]
ravitri: done

pkg/clusterversion/cv.go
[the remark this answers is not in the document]
ravitri: Done.. Actually, I see both the ways to be acceptable i.e, using an empty string or using the len.. But yes, used empty string for now..

pkg/upgraders/osd/upgrader.go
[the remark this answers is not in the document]
ravitri: This part of logging is moved to clusterversion package for now instead of dealing with it in the upgrader. That helps reduce some code. To be further improved as part of https://issues.redhat.com/browse/OSD-7686

pkg/validation/validation.go
```diff
@@ -57,131 +58,195 @@ const (
...
-		return ValidatorResult{
-			IsValid:           false,
-			IsAvailableUpdate: false,
-			Message:           fmt.Sprintf("Failed to parse desired version %s as semver", dv),
-		}, nil
-	}
-	currentVersion, err := semver.Parse(version)
-	if err != nil {
+	desiredChannel := uC.Spec.Desired.Channel
+	if (len(image) == 0 && len(dv) == 0) || (len(image) > 0 && len(desiredChannel) > 0) || (len(image) > 0 && len(dv) == 0 && len(desiredChannel) == 0) {
```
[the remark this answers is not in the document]
ravitri: Done.. Used empty(), supportsImageUpgrade() and supportsVersionUpgrade() functions to make the code more readable.. This is to be further generalized and improved as part of https://issues.redhat.com/browse/OSD-7686

pkg/upgraders/osd/upgrader.go
ravitri: This part is removed from upgrade and moved to clusterversion package..

pkg/clusterversion/cv.go
ravitri: done

pkg/clusterversion/cv.go
ravitri: done

pkg/clusterversion/cv.go
ravitri: done

ravitri:
For now, made the code better based on the comments.. Thanks a lot for the review 👍🏼 The further code refactoring to be tracked as part of https://issues.redhat.com/browse/OSD-7686 once the image/version logic is settled across validation and clusterversion packages. This would be done after the image version check is automated and the version dependency on the image is removed.

deploy/crds/upgrade.managed.openshift.io_upgradeconfigs_crd.yaml
ravitri: No I haven't. Do I need to? The CSV for MUO doesn't have the CRD definition in it as such. Do you see any potential issues? The test upgrades with image reference have run fine with updated CRD locally.

mrbarge:
Tested this out with the latest commit and it's working well, I'm personally happy with the PR at this point. I'll let @dofinn or @bmeng do the lgtm duties if they're happy with the updates.

pkg/clusterversion/cv.go
```diff
@@ -64,44 +68,67 @@ func (c *clusterVersionClient) GetClusterVersion() (*configv1.ClusterVersion, er
...
+		if clusterVersion.Spec.Channel != desired.Channel {
+			logger.Info(fmt.Sprintf("Setting ClusterVersion to Channel %s Version %s", desired.Channel, desired.Version))
+			clusterVersion.Spec.Channel = desired.Channel
+			err = c.client.Update(context.TODO(), clusterVersion)
+			if err != nil {
+				return false, err
+			}
+
+			// Retrieve the updated version
+			clusterVersion, err = c.GetClusterVersion()
```
bmeng, line 91: what is the purpose for fetching the clusterversion here? do we need to verify the returned version also?

pkg/clusterversion/cv.go
```diff
@@ -64,44 +68,67 @@ func (c *clusterVersionClient) GetClusterVersion() (*configv1.ClusterVersion, er
...
-	}
 
-	clusterVersion.Spec.Overrides = []configv1.ComponentOverride{}
-	clusterVersion.Spec.DesiredUpdate = &configv1.Update{Version: uc.Spec.Desired.Version}
-	err = c.client.Update(context.TODO(), clusterVersion)
-	if err != nil {
-		return false, err
+	// If neither (version+channel) nor (image) is defined, throw error.
+	if (empty(desired.Version) && empty(desired.Channel)) && empty(desired.Image) {
+		return false, fmt.Errorf("need atleast (version+channel) or (image) defined in UpgradeConfig")
```
bmeng, line 131: atleast -> at least, or maybe change it to `either`?

pkg/clusterversion/cv.go
```diff
@@ -64,44 +68,67 @@ func (c *clusterVersionClient) GetClusterVersion() (*configv1.ClusterVersion, er
...
+				updateAvailable = true
+			}
+		}
+		if !updateAvailable {
+			return false, nil
 		}
 
-		// Retrieve the updated version
-		clusterVersion, err = c.GetClusterVersion()
+		clusterVersion.Spec.Overrides = []configv1.ComponentOverride{}
```
bmeng, line 108: This is to empty the contains exist in the spec.Overrides?

pkg/clusterversion/cv.go
```diff
@@ -64,44 +68,67 @@ func (c *clusterVersionClient) GetClusterVersion() (*configv1.ClusterVersion, er
...
-		return false, nil
-	}
 
-	clusterVersion.Spec.Overrides = []configv1.ComponentOverride{}
-	clusterVersion.Spec.DesiredUpdate = &configv1.Update{Version: uc.Spec.Desired.Version}
-	err = c.client.Update(context.TODO(), clusterVersion)
-	if err != nil {
-		return false, err
+	// If neither (version+channel) nor (image) is defined, throw error.
+	if (empty(desired.Version) && empty(desired.Channel)) && empty(desired.Image) {
```
bmeng, line 130: if all three fields have value, the version and channel will take priority?

pkg/clusterversion/cv.go
```diff
@@ -151,24 +178,37 @@ func (c *clusterVersionClient) HasUpgradeCompleted(cv *configv1.ClusterVersion,
...
+	if !empty(uc.Spec.Desired.Version) && !empty(uc.Spec.Desired.Channel) && empty(uc.Spec.Desired.Image) {
+		if !isEqualVersion(clusterVersion, uc) {
+			return false, nil
+		} else {
+			logger.Info(fmt.Sprintf("ClusterVersion is already set to Channel %s Version %s", uc.Spec.Desired.Channel, uc.Spec.Desired.Version))
+		}
+	}
+
+	// TODO: Remove version dependency after work on OSD-7609
+	if !empty(uc.Spec.Desired.Version) && !empty(uc.Spec.Desired.Image) && empty(uc.Spec.Desired.Channel) {
```
bmeng, line 206: what if all the version, channel and image specified?

pkg/clusterversion/cv.go
```diff
@@ -199,8 +239,13 @@ func GetCurrentVersion(clusterVersion *configv1.ClusterVersion) (string, error)
...
 	if len(gotVersion) == 0 {
-		return gotVersion, fmt.Errorf("Failed to get current version")
+		return gotVersion, fmt.Errorf("failed to get current version")
 	}
 
 	return gotVersion, nil
 }
+
+// empty function checks if a given string is empty or not.
+func empty(s string) bool {
```
bmeng, line 249: do we need a separated func to check the string empty? can be done by `if len(string) == 0`

pkg/clusterversion/cv.go
ravitri: This change was introduced in https://github.com/openshift/managed-upgrade-operator/pull/168/files#diff-d7acacac799c3b89ebeedaaf44a111ed1cde3e3e09c3f333f67d25aa213fa96e actually.. The diff seems to show this change as part of a new block though.. I think this is just intended to fetch the updated `ClusterVersion` resource and no need to verify the version yet since that's replaced in code below as such..

pkg/clusterversion/cv.go
ravitri: Change introduced in https://github.com/openshift/managed-upgrade-operator/pull/168/files#diff-d7acacac799c3b89ebeedaaf44a111ed1cde3e3e09c3f333f67d25aa213fa96e but shown in diff as part of new block. I think the intent here is to nullify any overrides if they exist and make sure that all components are upgraded/created for the new version.

pkg/clusterversion/cv.go
ravitri: For now, not considering priority anywhere. The `validation` package currently will check that ONLY 1 of (image+version) or (channel+version) is mentioned. This PR would check that in the validation package in pkg/validation/validation.go. I tested it. If all three are mentioned, the validation will throw error.

pkg/clusterversion/cv.go
ravitri: If all three mentioned that validation pkg will throw error in the beginning itself before proceeding with upgrade. However, this block would only focus on checking if upgrade already commenced if just image+version is specified. Do you think we still need that check in `clusterversion` package? I think that might be dead code which won't be reached at all because `validation` would handle it before.

pkg/clusterversion/cv.go
ravitri: Ack.. Will change to `either`

pkg/clusterversion/cv.go
ravitri: Yes, but based on previous review comments like https://github.com/openshift/managed-upgrade-operator/pull/253#discussion_r659426643, the code seemed more readable so thought of putting it in empty(). A part of making this better would also be taken up in https://issues.redhat.com/browse/OSD-7686. Let me know what you think.

pkg/clusterversion/cv.go
bmeng: ah, after a bit more reading, I think I got this, it is just re-write the current clusterversion value (after the update above) to the `clusterVersion` variable.

pkg/clusterversion/cv.go
ravitri: Done

PR #256, opened 2021-07-02
Send notification when the scale up node timeout, and continue the upgrade

bmeng:
### What type of PR is this?
_feature_

As upgrade best practice, we would like to proceed the upgrade event if when the scale extra node for upgrade timeout. And notify CU about this.

_Fixes OSD-7108_

### DIFF
--- a/pkg/eventmanager/eventmanager.go
+++ b/pkg/eventmanager/eventmanager.go
@@ -30,7 +30,7 @@ const (
 	// UPGRADE_SCALE_DELAY_DESC describes the upgrade scaling delayed
 	UPGRADE_SCALE_DELAY_DESC = "Cluster upgrade to version %s is experiencing a delay attempting to scale up an additional worker node. The upgrade will continue to retry. This is an informational notification and no action is required by you."
 	// UPGRADE_SCALE_DELAY_SKIP_DESC describes the upgrade scaling skipped after delay
-	UPGRADE_SCALE_DELAY_SKIP_DESC = "Cluster upgrade to version %s has experienced issue during capacity reservation efforts. This could be caused by the AWS account service quota limitation, or the temporary connectivity issue from the new worker node to the masters or the external services. The upgrade will continue without extra compute. This is an informational notification and no action is required by you."
+	UPGRADE_SCALE_DELAY_SKIP_DESC = "Cluster upgrade to version %s has experienced an issue during capacity reservation efforts. This could be caused by AWS account service quota limitations or temporary connectivity issues to/from the new worker node. The upgrade will continue without extra compute. This is an informational notification and no action is required by you."
 )
 
 // EventManager enables implementation of an EventManager
--- a/pkg/eventmanager/eventmanager.go
+++ b/pkg/eventmanager/eventmanager.go
@@ -29,12 +29,14 @@ const (
 	UPGRADE_EXTDEPCHECK_DELAY_DESC = "Cluster upgrade to version %s is experiencing a delay as an external dependency of the upgrade is currently unavailable. The upgrade will continue to retry. This is an informational notification and no action is required by you."
 	// UPGRADE_SCALE_DELAY_DESC describes the upgrade scaling delayed
 	UPGRADE_SCALE_DELAY_DESC = "Cluster upgrade to version %s is experiencing a delay attempting to scale up an additional worker node. The upgrade will continue to retry. This is an informational notification and no action is required by you."
+	// UPGRADE_SCALE_DELAY_SKIP_DESC describes the upgrade scaling skipped after delay
+	UPGRADE_SCALE_DELAY_SKIP_DESC = "Cluster upgrade to version %s has experienced issue during capacity reservation efforts. This could be caused by the AWS account service quota limitation, or the temporary connectivity issue from the new worker node to the masters or the external services. The upgrade will continue without extra compute. This is an informational notification and no action is required by you."
 )
 
 // EventManager enables implementation of an EventManager
 //go:generate mockgen -destination=mocks/eventmanager.go -package=mocks github.com/openshift/managed-upgrade-operator/pkg/eventmanager EventManager
 type EventManager interface {
-	Notify(state notifier.NotifyState) error
+	Notify(state notifier.MuoState) error
 }
 
 // EventManagerBuilder enables implementation of an EventManagerBuilder
@@ -83,7 +85,7 @@ func (emb *eventManagerBuilder) NewManager(client client.Client) (EventManager, 
 	}, nil
 }
 
-func (s *eventManager) Notify(state notifier.NotifyState) error {
+func (s *eventManager) Notify(state notifier.MuoState) error {
 	// Get the current UpgradeConfig
 	uc, err := s.upgradeConfigManager.Get()
 	if err != nil {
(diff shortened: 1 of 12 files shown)


pkg/eventmanager/eventmanager.go
```diff
@@ -29,12 +29,14 @@ const (
 	UPGRADE_EXTDEPCHECK_DELAY_DESC = "Cluster upgrade to version %s is experiencing a delay as an external dependency of the upgrade is currently unavailable. The upgrade will continue to retry. This is an informational notification and no action is required by you."
 	// UPGRADE_SCALE_DELAY_DESC describes the upgrade scaling delayed
 	UPGRADE_SCALE_DELAY_DESC = "Cluster upgrade to version %s is experiencing a delay attempting to scale up an additional worker node. The upgrade will continue to retry. This is an informational notification and no action is required by you."
+	// UPGRADE_SCALE_DELAY_SKIP_DESC describes the upgrade scaling skipped after delay
+	UPGRADE_SCALE_DELAY_SKIP_DESC = "Cluster upgrade to version %s is experiencing a delay attempting to scale up an additional worker node. The upgrade will continue without scaling up extra nodes. This is an informational notification and no action is required by you."
```
bmeng, line 33: @wgordon17 Could you help check if the message here is proper?

dofinn:
```suggestion
	UPGRADE_SCALE_DELAY_SKIP_DESC = "Cluster upgrade to version %s has experienced an unforseen issue during capacity reservation efforts. The upgrade will continue without extra compute. This is an informational notification and no action is required by you."
```

dofinn:
Please unhold when ready to merge after agreed wording on delayed message.

pkg/notifier/ocmnotifier.go
```diff
@@ -95,46 +119,48 @@ func (s *ocmNotifier) getPolicyIdForUpgradeConfig(clusterId string) (*string, er
...
+	case OcmStateDelayed:
 		// can go to completed or failed state
 		switch to {
-		case StateCompleted:
+		case OcmStateCompleted:
+			return true
+		case OcmStateFailed:
 			return true
-		case StateFailed:
+		case OcmStateDelayed:
```
bmeng, line 155: Have to add this here, since it will change from a delay to another delay.
The problem here is if we do delayerTrigger and the scaleTimeout have different values, it will send the delay notification every reconcile loop.

mrbarge: I agree this is a problem, and I've been able to replicate it not just with scale timeout logs but with delays due to health check whilst testing the PR.

Maybe we can rework this function to be based around `MuoState` rather than `OcmState`?

It would be similar rules except:

- a MuoStateDelayed can transition to a MuoStateDelayed but cannot transition to a MuoStateDelayed
- a MuoStateSkiped can transition to a MuoStateDelayed but cannot transition to MuoStateSkipped.

That doesn't solve 100% of the issues (can still flap between delayed -> skipped and skipped -> delayed) but might cut down on at least some of the potential for repetition.

bmeng: In that case, maybe we can just disable the transition from skipped to delay?

But there is one problem for this, when sending the notification in `func (s *ocmNotifier) NotifyState`, we need to fetch the current OCM state. Then we need to convert this back to Muo state.

The problem is the state map has two values matched to `OcmStateDelayed`, I think if we write a new function to convert the OCM state to Muo state, it will return a random value from MuoStateSkipped and MuoStateDelayed.

WDYT?

pkg/upgraders/osd/upgrader.go
```diff
@@ -403,7 +409,8 @@ func ControlPlaneUpgraded(c client.Client, cfg *osdUpgradeConfig, scaler scaler.
 
 // SendStartedNotification sends a notification on upgrade commencement
 func SendStartedNotification(c client.Client, cfg *osdUpgradeConfig, scaler scaler.Scaler, dsb drain.NodeDrainStrategyBuilder, metricsClient metrics.Metrics, m maintenance.Maintenance, cvClient cv.ClusterVersion, nc eventmanager.EventManager, upgradeConfig *upgradev1alpha1.UpgradeConfig, machinery machinery.Machinery, availabilityCheckers ac.AvailabilityCheckers, logger logr.Logger) (bool, error) {
-	err := nc.Notify(notifier.StateStarted)
+	err := nc.Notify(notifier.MuoStateStarted)
+	fmt.Println("send started")
```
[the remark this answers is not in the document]
bmeng: Ah, thanks. forget to remove thoes debug strings...

mrbarge:
Successfully tested in Staging by testing a scenario where a cluster is first delayed due to External Dependency check failure, and then delayed again due to Scale-up failure, verified that the upgrade proceeded regardless and that correct service logs were sent.

/lgtm

Feel free to unhold when you've got the go-ahead re messaging (and kindly let me know so I can port impacted bits to the refactor PR)

bmeng:
@wgordon17 could you help check if the message added here as UPGRADE_SCALE_DELAY_SKIP_DESC to send to customer via service log when the scaling timeout and going to skip is proper?

pkg/eventmanager/eventmanager.go
```diff
@@ -29,12 +29,14 @@ const (
 	UPGRADE_EXTDEPCHECK_DELAY_DESC = "Cluster upgrade to version %s is experiencing a delay as an external dependency of the upgrade is currently unavailable. The upgrade will continue to retry. This is an informational notification and no action is required by you."
 	// UPGRADE_SCALE_DELAY_DESC describes the upgrade scaling delayed
 	UPGRADE_SCALE_DELAY_DESC = "Cluster upgrade to version %s is experiencing a delay attempting to scale up an additional worker node. The upgrade will continue to retry. This is an informational notification and no action is required by you."
+	// UPGRADE_SCALE_DELAY_SKIP_DESC describes the upgrade scaling skipped after delay
+	UPGRADE_SCALE_DELAY_SKIP_DESC = "Cluster upgrade to version %s has experienced an unforeseen issue during capacity reservation efforts. The upgrade will continue without extra compute. This is an informational notification and no action is required by you."
```
wgordon17, line 33: Is this truly an unforeseen/unexpected issue? What types of scenarios is it possible that we could run into this issue? Saying that something is "unforeseen" makes it appear that we were lazy or not thorough enough. I'd just like to be as explicit as possible.

bmeng: Hi @wgordon17 was trying to implement the dynamic message by reading the failure status from the machine. But turns out that the MachineHealthCheck currently deployed in our cluster will remove the machine as well as the machine status.
Based on our discussion, the reason could be AWS service quota issue, temporary node connectivity issue to the master or to the external dependency services, like quay. 

Would you like to review the new message? Thanks!

pkg/eventmanager/eventmanager.go
```diff
@@ -29,12 +29,14 @@ const (
 	UPGRADE_EXTDEPCHECK_DELAY_DESC = "Cluster upgrade to version %s is experiencing a delay as an external dependency of the upgrade is currently unavailable. The upgrade will continue to retry. This is an informational notification and no action is required by you."
 	// UPGRADE_SCALE_DELAY_DESC describes the upgrade scaling delayed
 	UPGRADE_SCALE_DELAY_DESC = "Cluster upgrade to version %s is experiencing a delay attempting to scale up an additional worker node. The upgrade will continue to retry. This is an informational notification and no action is required by you."
+	// UPGRADE_SCALE_DELAY_SKIP_DESC describes the upgrade scaling skipped after delay
+	UPGRADE_SCALE_DELAY_SKIP_DESC = "Cluster upgrade to version %s has experienced issue during capacity reservation efforts. This could be caused by the AWS account service quota limitation, or the temporary connectivity issue from the new worker node to the masters or the external services. The upgrade will continue without extra compute. This is an informational notification and no action is required by you."
```
mrbarge, line 33: Made some slight tweaks - a couple of tiny grammar things and just simplifying the issues around the worker.

```suggestion
	UPGRADE_SCALE_DELAY_SKIP_DESC = "Cluster upgrade to version %s has experienced an issue during capacity reservation efforts. This could be caused by AWS account service quota limitations or temporary connectivity issues to/from the new worker node. The upgrade will continue without extra compute. This is an informational notification and no action is required by you."
```

pkg/eventmanager/eventmanager.go
bmeng: @spurtell Could you help take a look here?

PR #262, opened 2021-07-19
Remove version dependency for image digest upgrades

ravitri:
### What type of PR is this?
_(cleanup/refactor)_

This PR would remove the dependency on providing `version` with `image` during MUO upgrades. Thus now, the end user may or may not provide an associated `version` when mentioning an `image` digest in the UpgradeConfig. 

The initial validation may or may not have the `version` initially but the UpgradeConfig will always be updated with a corresponding `version` for MUO to refer but will eventually only update the `image` in the ClusterVersion resource for CVO.

_Fixes #_ https://issues.redhat.com/browse/OSD-7609

1. Currently the update of `version` in UpgradeConfig is done in the validation package itself since the image and version comparison and update could be done together as part of validation. If required, in future the update part can be handled elsewhere.
2. Image based and version based upgrades ran fine so far. 
3. Metrics/alerting wasn't changed so far but can make changes if necessary.

- [x] Added test corrections

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,5 +82,4 @@ spec:
   PDBForceDrainTimeout: 60
   desired:
     image: "quay.io/openshift-release-dev/ocp-release@sha256:783a2c963f35ccab38e82e6a8c7fa954c3a4551e07d2f43c06098828dd986ed4"
-    version: "4.7.13"
 ```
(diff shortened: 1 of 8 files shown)


ravitri:
/hold
Covering an edge case for this PR.

PR #259, opened 2021-07-13
Update boilerplate & update crd to v1

sam-nguyen7:
### What type of PR is this?
Feature

### What this PR does / why we need it?
Update boilerplate version from v1.0 to v2.1
Update crd version from v1beta1 to v1

### Which Jira/Github issue(s) this PR fixes?
https://issues.redhat.com/browse/OSD-7597

### Pre-checks (if applicable):
- [ ] Tested latest changes against a cluster
- [x] Ran `make generate` command locally to validate code changes
- [ ] Included documentation changes with PR

PR #263, opened 2021-07-26
Adding controller-gen tool as part of MUO tools

ravitri:
### What type of PR is this?
_(bug)_

As part of PR https://github.com/openshift/managed-upgrade-operator/pull/259, the `make generate` command will fail locally if there is no `controller-gen` binary present. Refer:
```bash
$ make generate
# The artist formerly known as `operator-sdk generate crds`:
cd pkg/apis; controller-gen crd:crdVersions=v1 paths=./... output:dir=../../deploy/crds
/bin/sh: line 1: controller-gen: command not found
make: *** [boilerplate/openshift/golang-osd-operator/standard.mk:157: op-generate] Error 127
```
This PR would add the dependency such that `controller-gen` is fetched when `make tools` is run.

- [X] Tested latest changes against a cluster
- [X] Ran `make generate` command locally to validate code changes

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -34,6 +34,7 @@ require (
 	k8s.io/client-go v12.0.0+incompatible
 	k8s.io/kube-openapi v0.0.0-20201113171705-d219536bb9fd
 	sigs.k8s.io/controller-runtime v0.8.3
+	sigs.k8s.io/controller-tools v0.3.0
 )
 
 replace (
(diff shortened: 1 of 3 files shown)


2uasimojo:
The preferred way is to run boilerplate-provided targets under `container-make`, e.g.

```
$ ./boilerplate/_lib/container-make generate
```

That way you know you're getting the _right_ version of `controller-gen`, and any other tools being used.

PR #267, opened 2021-08-03
upgradeconfig validation refactor

bmeng:
### What type of PR is this?
_(refactor)_

remove the redundant in validation and upgrade

_Fixes #OSD-7686_

### DIFF
--- a/pkg/clusterversion/cv.go
+++ b/pkg/clusterversion/cv.go
@@ -78,8 +78,6 @@ func (c *clusterVersionClient) EnsureDesiredConfig(uc *upgradev1alpha1.UpgradeCo
 		return false, err
 	}
 
-	desired := uc.Spec.Desired
-
 	// Check which upgrade spec source we are going to use
 	upgradeSource, err := checkUpgradeSource(uc)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/clusterversion/cv.go
```diff
@@ -111,26 +132,10 @@ func (c *clusterVersionClient) EnsureDesiredConfig(uc *upgradev1alpha1.UpgradeCo
...
-		}
-	}
-
-	// If neither (version+channel) nor (image) is defined, throw error.
-	if (empty(desired.Version) && empty(desired.Channel)) && empty(desired.Image) {
-		return false, fmt.Errorf("need either (version+channel) or (image) defined in UpgradeConfig")
-	}
-
-	return true, nil
+	return false, nil
```
dofinn, line 138: This occurs when a non valid upgradesource is present? we should throw error here?

bmeng: updated to print error message when it returns false, also fixed several other places with the same `false, nil` condition

mrbarge:
/hold
This has an issue that needs fixing, will discuss.

mrbarge:
/unhold
Thanks for sorting that issue :)  Will continue review.

pkg/clusterversion/cv.go
```diff
@@ -76,9 +80,26 @@ func (c *clusterVersionClient) EnsureDesiredConfig(uc *upgradev1alpha1.UpgradeCo
...
+		if clusterVersion.Spec.DesiredUpdate == nil {
+			logger.Info(fmt.Sprintf("Setting ClusterVersion to Image %s", desired.Image))
+			clusterVersion.Spec.DesiredUpdate = &configv1.Update{Image: desired.Image}
+			err = c.client.Update(context.TODO(), clusterVersion)
+			if err != nil {
+				return false, err
+			}
+			return true, nil
+		}
+		return false, fmt.Errorf("clusterversion already had desiredUpdate set, might be pervious upgrade wasn't completed successfully")
```
mrbarge, line 100: This section also has an issue in that it will prevent any further image-based upgrades from occuring once the cluster has upgraded once.

When a cluster finishes upgrading, `desiredUpdate` remains in place in `clusterversion`. So once a cluster has upgraded once, this will never be `nil`. A second upgrade will cause it to drop down to this error path.

bmeng: hmm, I removed the `clusterVersion.Spec.DesiredUpdate.Image != desired.Image` from the original code, but did not test with the second upgrade. Thanks for catch this.

mrbarge:
Nice tidy-up here.
/lgtm

PR #268, opened 2021-08-03
OSD-7557 Use Services over Routes for prom/AM

dofinn:
### What type of PR is this?
Refactor/Feature

The use of routes required the cluster to make egress calls. This is not desirable when a firewall preventing egress access is being employed for security. 

The use of internal cluster services requires the retrieval of the monitoring services self signed certificate authority to avoid TLS authority errors. This PR gets the `CA` from the `openshift-monitoring` namespace and adds it to both the prometheus and alertmanager transports so secure and error free communication can occur with these internal cluster services. 

* Moved MUO config specific data to `config/config.go` rather then remain in the implementation of the `configmanager` pkg.
* refactored `configmanager` to receive a `target` that provides defualts rather then a namespace string
* Added `port-forwards` script to aid in new local dev requirements
* updated development doc with run modes

### Extra Resources
https://kubernetes.io/docs/tasks/tls/managing-tls-in-a-cluster/#trusting-tls-in-a-cluster 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,6 @@ run-standard-routes: ## Run locally with openshift-managed-upgrade-operator as O
 tools: ## Install local go tools for MUO
 	cat tools.go | grep _ | awk -F'"' '{print $$2}' | xargs -tI % go install %
 
-.DEFAULT_GOAL := help
 .PHONY: help
 help: ## Show this help screen.
 		@echo 'Usage: make <OPTIONS> ... <TARGETS>'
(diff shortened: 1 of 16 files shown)


bmeng:
there are two separated changes (configmanager and prometheus route/service) put together?

dofinn:
> there are two separated changes (configmanager and prometheus route/service) put together?

Yes, I originally I was going to use to the configmanager to retrieve the configmap however it has a hard requirement on the .data payload being yaml. I realised this after refactoring :) I can split this in two if its more consumable.

dofinn:
Breaking into smaller PRs. 
Pt 1: https://github.com/openshift/managed-upgrade-operator/pull/271

PR #271, opened 2021-08-05
OSD-7557 Part 1 configmanager refactor 

dofinn:
### What type of PR is this?
refactor

This was a byproduct of a following PR. Enables a more expansible implementation of configmanager. 

* Moved MUO config specific data to `config/config.go` rather then remain in the implementation of the `configmanager` pkg.
* refactored `configmanager` to receive a `target` that provides defualts rather then a namespace string

- [] Ran `make generate` command locally to validate code changes

### DIFF
--- a/pkg/controller/nodekeeper/nodekeeper_controller.go
+++ b/pkg/controller/nodekeeper/nodekeeper_controller.go
@@ -5,7 +5,7 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/openshift/managed-upgrade-operator/util"
+	"github.com/openshift/managed-upgrade-operator/config"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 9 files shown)


pkg/controller/nodekeeper/nodekeeper_controller.go
```diff
@@ -142,7 +143,11 @@ func (r *ReconcileNodeKeeper) Reconcile(ctx context.Context, request reconcile.R
 	if err != nil {
 		return reconcile.Result{}, nil
 	}
-	cfm := r.configManagerBuilder.New(r.client, operatorNamespace)
+
+	target := config.CMTarget{Namespace: operatorNamespace}
```
bmeng, line 147: I see we have multiple places with `util.GetOperatorNamespace()`
And the put the value to the CMTarget.

Can we just move the `util.GetOperatorNamespace()` to `config/config.go`? so that we can call the `NewCMTarget()` directly.

dofinn: Move it so we call `config.GetOperatorNamespace()` ?

bmeng: move the `util.GetOperatorNamespace()` inside the `NewCMTarget()` so that we do not need to call it separately

dofinn: what if we want to pass an explicit namespace in future?

bmeng: I think it should still be able to read from args as what it current do? just set the default from the return of `util.GetOperatorNamespace()` instead of the hardcoded `OperatorNamespace`

mrbarge:
:+1: Seems fine to me.

pkg/controller/nodekeeper/nodekeeper_controller.go
dofinn: sounds good. Resolved, squashed and pushed. Please review.

PR #273, opened 2021-08-10
pt 2 route 2 service

dofinn:
### What type of PR is this?
Refactor/Feature

The use of routes required the cluster to make egress calls. This is not desirable when a firewall preventing egress access is being employed for security. 

The use of internal cluster services requires the retrieval of the monitoring services self signed certificate authority to avoid TLS authority errors. This PR gets the `CA` from the `openshift-monitoring` namespace and adds it to both the prometheus and alertmanager transports so secure and error free communication can occur with these internal cluster services. 

* Added `port-forwards` script to aid in new local dev requirements
* updated development doc with run modes

### Extra Resources
https://kubernetes.io/docs/tasks/tls/managing-tls-in-a-cluster/#trusting-tls-in-a-cluster 


docs/development.md
```diff
@@ -99,30 +124,100 @@ Regardless of how you choose to run the operator, before doing so ensure the `Up
...
+
+
 ### Locally
 
 - Make sure you have the [operator-sdk](#operator-sdk) `$PATH`.
 
-- If you are not using an account that has `cluster-admin` privileges, you will need to [elevate permissions](https://github.com/openshift/ops-sop/blob/master/v4/knowledge_base/manage-privileges.md) to possess them.
+- You will need to be logged in with an account that meets the [RBAC requirements](https://github.com/openshift/managed-upgrade-operator/blob/master/deploy/cluster_role.yaml) for the MUO service account.
 
+<<<<<<< Updated upstream
```
bmeng, line 151: git rebase residual here and some others in this file

development/port-forwards
```diff
@@ -0,0 +1,37 @@
...
+
+# Setup for local resolution
+grep 'prometheus' /etc/hosts > /dev/null
+if [[ $? == 1 ]]
+then
+	sudo -- sh -c "echo 127.0.0.1 prometheus-k8s.openshift-monitoring.svc.cluster.local alertmanager-main.openshift-monitoring.svc.cluster.local >> /etc/hosts"
+fi
+
+# Setup prometheus and alertmanager port-forwards
+while true; do $OC port-forward -n openshift-monitoring svc/prometheus-k8s 9091:9091;done &
```
bmeng, line 36: why need to run as a loop here?

pkg/metrics/metrics.go
```diff
@@ -351,20 +405,66 @@ func (c *Counter) IsAlertFiring(alert string, checkedNS, ignoredNS []string) (bo
 	return false, nil
 }
 
-func getPromHost(c client.Client) (*string, error) {
+// GetServiceEndpoint accepts a client,namespace,svcName and portName and attempts to retrive
+// the services endpoint in the form of resolveable.service:portnumber.
+func GetServiceEndpoint(c client.Client, namespace, svcName, portName string) (string, error) {
```
bmeng, line 410: The ServiceEndpoint here is actually not an endpoint resource right? it is just the service+port. The function name is a bit confusing.

And I think it can be handled by `const` instead of function.

pkg/metrics/metrics.go
```diff
@@ -78,41 +90,83 @@ func NewBuilder() MetricsBuilder {
...
 		},
 	}, nil
 }
 
 type prometheusRoundTripper struct {
 	token string
+	tls   *tls.Config
+}
+
+func TLSConfig(c client.Client) (*tls.Config, error) {
```
bmeng, line 129: Should better explain the function if it is public

pkg/metrics/metrics.go
```diff
@@ -351,20 +405,66 @@ func (c *Counter) IsAlertFiring(alert string, checkedNS, ignoredNS []string) (bo
...
+		endpoint, err = getRouteEndpoint(c, appName)
+		if err != nil {
+			return endpoint, err
+		}
+	}
+
+	return endpoint, nil
+}
+
+func getRouteEndpoint(c client.Client, appName string) (string, error) {
```
bmeng, line 454: function name to `getRouteHostName` might be less confusing

pkg/metrics/metrics.go
```diff
@@ -351,20 +405,66 @@ func (c *Counter) IsAlertFiring(alert string, checkedNS, ignoredNS []string) (bo
...
+}
+
+func Endpoint(c client.Client, namespace, appName, portName string) (string, error) {
+	var endpoint string
+	var err error
+
+	runLocal := isRunModeLocal()
+	useRoutes := config.UseRoutes()
+
+	if !runLocal || runLocal && !useRoutes {
```
bmeng, line 439: redundant logic of `!runLocal || runLocal`?

pkg/metrics/metrics.go
```diff
@@ -351,20 +405,66 @@ func (c *Counter) IsAlertFiring(alert string, checkedNS, ignoredNS []string) (bo
...
+	}
+	endpoint := fmt.Sprint(host + ":" + port)
+	return endpoint, nil
+}
+
+func isRunModeLocal() bool {
+	return os.Getenv(k8sutil.ForceRunModeEnv) == string(k8sutil.LocalRunMode)
+}
+
+func Endpoint(c client.Client, namespace, appName, portName string) (string, error) {
```
bmeng, line 432: public func needs comments.

And name it like `ServiceCanonicalName` would be less confusing

pkg/metrics/metrics.go
```diff
@@ -351,20 +405,66 @@ func (c *Counter) IsAlertFiring(alert string, checkedNS, ignoredNS []string) (bo
...
+	}
+
+	host := fmt.Sprintf(svcName + "." + namespace + clusterSVCSuffix)
+	var port string
+	for _, p := range svc.Spec.Ports {
+		if p.Name == portName {
+			port = strconv.FormatInt(int64(p.Port), 10)
+		}
+	}
+	endpoint := fmt.Sprint(host + ":" + port)
```
mrbarge, line 424: Add some validation / error returning here if port remains an empty string having not been set.

docs/development.md
dofinn: Thanks @bmeng this not appear when i rebased :0

development/port-forwards
dofinn: The connections are not stable. If they are not ran like this the upgrade will be disrupted.

pkg/metrics/metrics.go
dofinn: This function name is confusing in the context of kubernetes. I will update it to `GetService`. I think we should keep as function rather then consts in-case ports change in future.

pkg/metrics/metrics.go
dofinn: Done, also updated to MonitoringTLSConfig

pkg/metrics/metrics.go
dofinn: updated.

pkg/metrics/metrics.go
dofinn: No
If `!runLocal` = its not running local so its in production (or atleast a cluster)
if `runLocal && !useRoutes` = its running locally (`make run-standard) and I want to not use routes and use internal services of the cluster

pkg/metrics/metrics.go
dofinn: Terminology here is hard in kubernetes as objects exist for endpoints,services and routes. As this function can return both a service or route `host:port` I am renaming the function `NetworkTarget`.

pkg/metrics/metrics.go
dofinn: added validation. why is `Sprint()` redundant?

mrbarge: Thanks for fixing, looks good.
re the `Sprint()` I meant it could be
`endpoint := host + ":" + port`
but that's a trivial matter.

pkg/metrics/metrics.go
```diff
@@ -351,20 +407,73 @@ func (c *Counter) IsAlertFiring(alert string, checkedNS, ignoredNS []string) (bo
 	return false, nil
 }
 
-func getPromHost(c client.Client) (*string, error) {
+// GetService accepts a client,namespace,svcName and portName and attempts to retrive
+// the services endpoint in the form of resolveable.service:portnumber.
+func GetService(c client.Client, namespace, svcName, portName string) (string, error) {
```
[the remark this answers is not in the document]
dofinn: why not build it programatically to account for any variable changes?

pkg/metrics/metrics.go
bmeng: I am still not quite sure, the current logic is 
`if !A || A && !B {}`

and the result would be:
```
t t > no call

t f > call

f t > call

f f > call 
```

And if we use the logic like below, it will have the same behaviour?
`if !A || !B {}`

dofinn: how do i `runLocal` and `useRoutes` then?

pkg/metrics/metrics.go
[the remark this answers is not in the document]
bmeng: In the current implementation, it is still depends on the service name which defined in variable like `alertmanager-main` and `web`. I cannot see much benefit with function but one more api call here.

PR #277, opened 2021-09-13
CRD validation for PDBForceDrainTimeout field

a7vicky:
### What type of PR is this?
refactor

CRD validation for  PDBForceDrainTimeout field in the upgradeconfig CR. It will throw validation error if we put negative integer value for this parameter.

[OSD-7898](https://issues.redhat.com/browse/OSD-7898)

### DIFF
--- a/pkg/apis/upgrade/v1alpha1/upgradeconfig_types.go
+++ b/pkg/apis/upgrade/v1alpha1/upgradeconfig_types.go
@@ -25,6 +25,7 @@ type UpgradeConfigSpec struct {
 	// Specify the upgrade start time
 	UpgradeAt string `json:"upgradeAt"`
 
+	// +kubebuilder:validation:Minimum:=0
 	// The maximum grace period granted to a node whose drain is blocked by a Pod Disruption Budget, before that drain is forced. Measured in minutes.
 	PDBForceDrainTimeout int32 `json:"PDBForceDrainTimeout"`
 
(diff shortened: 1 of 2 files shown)


pkg/apis/upgrade/v1alpha1/upgradeconfig_types.go
```diff
@@ -25,6 +25,7 @@ type UpgradeConfigSpec struct {
 	// Specify the upgrade start time
 	UpgradeAt string `json:"upgradeAt"`
 
+	// +kubebuilder:validation:Minimum:=0
 	// The maximum grace period granted to a node whose drain is blocked by a Pod Disruption Budget, before that drain is forced. Measured in minutes.
```
mrbarge, line 29: I have one small request, could you also update the comment description here to state what a '0' means in the context of this field and how MUO behaves? So that we don't have to wonder in the future what MUO does with a '0' value? :)  (if my memory is correct from your previous testing that meant that MUO would grant no grace period at all)

a7vicky: Thanks for the review @mrbarge  The minimum value here to just restrict and not allowing any negative integer value for this param. Also "0" value is currently not respecting `nodeDrain.expectedNodeDrainTime` set in the configmap and executing pdb delete strategy immediately. This issue is tracked here [OSD-7899](https://issues.redhat.com/browse/OSD-7899).
Still i am not clear that it should respect nodeDrain.expectedNodeDrainTime if PDBForceDrainTimeout<= nodeDrain.expectedNodeDrainTime Or just for "0" value.

mrbarge: Oh yeah, understood - to clarify my request, I meant updating the actual field comment in the code of what "0" means in terms of behaviour (so that it also gets reflected in the CRD field description  :)  ). 
The change looks good to me otherwise.

mrbarge: Thanks for the change :)
/lgtm

PR #283, opened 2021-10-14
[OSD-7882] Move getPodList functions out of drain strategies into pod…

T0MASD:
… package

### What type of PR is this?
_(refactor)_

Apply DRY to PodList functions in drain strategies

_Fixes #_ [OSD-7882](https://issues.redhat.com/browse/OSD-7882)

- [x] Ran `make test` command locally to validate code changes

### DIFF
--- a/pkg/drain/podDeleteStrategy_test.go
+++ b/pkg/drain/podDeleteStrategy_test.go
@@ -155,23 +155,4 @@ var _ = Describe("Pod Delete Strategy", func() {
 
 		})
 	})
-
-	Context("Get Pod List to be deleted", func() {
-		It("Returns list of pods with no errors", func() {
-			gomock.InOrder(
-				mockKubeClient.EXPECT().List(gomock.Any(), gomock.Any(), gomock.Any()).SetArg(1, podList),
-			)
-			_, err := pod.GetPodList(pds.client, node, pds.filters)
-			Expect(err).To(BeNil())
-		})
-
-		It("Returns no pods if there is any error while listing pods", func() {
-			gomock.InOrder(
-				mockKubeClient.EXPECT().List(gomock.Any(), gomock.Any(), gomock.Any()).SetArg(1, podList).Return(fmt.Errorf("fake error")),
-			)
-			_, err := pod.GetPodList(pds.client, node, pds.filters)
-			Expect(err).To(HaveOccurred())
-			Expect(err).NotTo(BeNil())
-		})
-	})
 })
--- a/pkg/drain/removeFinalizersStrategy_test.go
+++ b/pkg/drain/removeFinalizersStrategy_test.go
@@ -156,25 +156,4 @@ var _ = Describe("Remove Finalizer Strategy", func() {
 
 		})
 	})
-
-	Context("Get Pod List with finalizers", func() {
-		It("Returns list of pods with no errors", func() {
-			gomock.InOrder(
-				mockKubeClient.EXPECT().List(gomock.Any(), gomock.Any(), gomock.Any()).SetArg(1, podList),
-			)
-			_, err := pod.GetPodList(rfs.client, node, rfs.filters)
-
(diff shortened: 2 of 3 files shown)


pkg/drain/podDeleteStrategy.go
```diff
@@ -16,7 +13,8 @@ type podDeletionStrategy struct {
 }
 
 func (pds *podDeletionStrategy) Execute(node *corev1.Node) (*DrainStrategyResult, error) {
-	podsToDelete, err := pds.getPodList(node)
+	filters := []pod.PodPredicate{isOnNode(node)}
```
mrbarge, line 16: Any reason why no appending of the podDeletionStrategy filters here? ie.
```suggestion
	filters := append([]pod.PodPredicate{isOnNode(node)}, pds.filters...)
```
That seems to be inadvertently being dropped through this change. I notice it is being done for the `stuckTerminatingStrategy.go` file, but not here.

pkg/drain/podDeleteStrategy.go
```diff
@@ -34,29 +32,11 @@ func (pds *podDeletionStrategy) Execute(node *corev1.Node) (*DrainStrategyResult
 }
 
 func (pds *podDeletionStrategy) IsValid(node *corev1.Node) (bool, error) {
-	targetPods, err := pds.getPodList(node)
+	filters := []pod.PodPredicate{isOnNode(node)}
```
mrbarge, line 35: Same question/comment from earlier applies here.

pkg/drain/removeFinalizersStrategy.go
```diff
@@ -16,7 +13,8 @@ type removeFinalizersStrategy struct {
 }
 
 func (rfs *removeFinalizersStrategy) Execute(node *corev1.Node) (*DrainStrategyResult, error) {
-	podsWithFinalizers, err := rfs.getPodList(node)
+	filters := []pod.PodPredicate{isOnNode(node), hasFinalizers}
```
mrbarge, line 16: Same question/comment from earlier applies here.

pkg/drain/removeFinalizersStrategy.go
```diff
@@ -33,29 +31,12 @@ func (rfs *removeFinalizersStrategy) Execute(node *corev1.Node) (*DrainStrategyR
 }
 
 func (rfs *removeFinalizersStrategy) IsValid(node *corev1.Node) (bool, error) {
-	targetPods, err := rfs.getPodList(node)
+	filters := []pod.PodPredicate{isOnNode(node), hasFinalizers}
```
mrbarge, line 34: Same question/comment from earlier applies here.

pkg/pod/pod.go
```diff
@@ -88,3 +89,22 @@ func RemoveFinalizersFromPod(c client.Client, pl *corev1.PodList) (*RemoveFinali
...
+
+	allPods := &corev1.PodList{}
+	err = c.List(context.TODO(), allPods, &client.ListOptions{
+		FieldSelector: fieldSelector,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return FilterPods(allPods, filters...), nil
```
bmeng, line 109: it returns a function here? why cannot return a slice of pods instead?

T0MASD: Do you mean this part should be updated to:
```
filteredPods := FilterPods(allPods, filters...), nil
return filteredPods
```
FWIW the old function was also returning the function:
```
return pod.FilterPods(allPods, filters...), nil
```
Functions are returned in other places as seen from
```
$ egrep -Rn "return \w+\(.+\)" pkg/ | wc -l
31
```

mrbarge: It seems to be returning the output of that function (`*corev1.PodList`) not the function itself. 
Which seems fine to me given that the function itself only returns one value.

PR #289, opened 2021-11-02
OSD-8404 : MUO to adhere CVO's Ack-Gate

anjoshi24:
### What type of PR is this?
Refactor

MUO needs to check CVO's ack gate condition before the upgrade and make decision based on the status.

https://issues.redhat.com/browse/OSD-8404
_Fixes #_

- [ ] Ran `make generate` command locally to validate code changes

### DIFF
--- a/pkg/eventmanager/eventmanager.go
+++ b/pkg/eventmanager/eventmanager.go
@@ -20,6 +20,8 @@ const (
 	UPGRADE_EXTDEPCHECK_FAILED_DESC = "Cluster upgrade to version %s was cancelled during the External Dependency Availability Check step. A required external dependency of the upgrade was unavailable, so the upgrade did not proceed. Automated upgrades will be retried on their next scheduling cycle. If you have manually scheduled an upgrade instead, it must now be rescheduled."
 	// UPGRADE_SCALE_FAILED_DESC describes the upgrade scaling failed
 	UPGRADE_SCALE_FAILED_DESC = "Cluster upgrade to version %s was cancelled during the Scale-Up Worker Node step. A temporary additional worker node was unable to be created to temporarily house workloads, so the upgrade did not proceed. Automated upgrades will be retried on their next scheduling cycle. If you have manually scheduled an upgrade instead, it must now be rescheduled."
+	// UPGRADE_UPGRADEABLE_FAILED_DESC describes the upgradeable check failure
+	UPGRADE_UPGRADEABLE_FAILED_DESC = "Cluster upgrade to version %s was cancelled because manual acknowledgement is required concerning the removal of deprecated APIs. Please review the Knowledge Base article at https://access.redhat.com/articles/6329921 and perform the acknowledgement procedure if you are not using workloads that require the deprecated APIs and wish to proceed with the upgrade. Automated upgrades will be retried on their next scheduling cycle. If you have manually scheduled an upgrade instead, it must be rescheduled."
 
 	// UPGRADE_DEFAULT_DELAY_DESC describes the upgrade default delay
 	UPGRADE_DEFAULT_DELAY_DESC = "Cluster upgrade to version %s is experiencing a delay whilst it performs necessary pre-upgrade procedures. The upgrade will continue to retry. This is an informational notification and no action is required."
@@ -164,6 +166,8 @@ func createFailureDescription(uc *v1alpha1.UpgradeConfig) string {
 	}
 
 	switch failedCondition.Type {
+	case v1alpha1.IsClusterUpgradable:
+		description = fmt.Sprintf(UPGRADE_UPGRADEABLE_FAILED_DESC, uc.Spec.Desired.Version)
 	case v1alpha1.UpgradePreHealthCheck:
 		description = fmt.Sprintf(UPGRADE_PREHEALTHCHECK_FAILED_DESC, uc.Spec.Desired.Version)
 	case v1alpha1.ExtDepAvailabilityCheck:
--- a/pkg/eventmanager/eventmanager.go
+++ b/pkg/eventmanager/eventmanager.go
@@ -20,8 +20,6 @@ const (
 	UPGRADE_EXTDEPCHECK_FAILED_DESC = "Cluster upgrade to version %s was cancelled during the External Dependency Availability Check step. A required external dependency of the upgrade was unavailable, so the upgrade did not proceed. Automated upgrades will be retried on their next scheduling cycle. If you have manually scheduled an upgrade instead, it must now be rescheduled."
 	// UPGRADE_SCALE_FAILED_DESC describes the upgrade scaling failed
 	UPGRADE_SCALE_FAILED_DESC = "Cluster upgrade to version %s was cancelled during the Scale-Up Worker Node step. A temporary additional worker node was unable to be created to temporarily house workloads, so the upgrade did not proceed. Automated upgrades will be retried on their next scheduling cycle. If you have manually scheduled an upgrade instead, it must now be rescheduled."
-	// UPGRADE_UPGRADEABLE_FAILED_DESC describes the upgradeable check failure
-	UPGRADE_UPGRADEABLE_FAILED_DESC = "Cluster upgrade to version %s was cancelled because manual acknowledgement is required concerning the removal of deprecated APIs. Please review the Knowledge Base article at https://access.redhat.com/articles/6329921 and perform the acknowledgement procedure if you are not using workloads that require the deprecated APIs and wish to proceed with the upgrade. Automated upgrades will be retried on their next scheduling cycle. If you have manually scheduled an upgrade instead, it must be rescheduled."
 
 	// UPGRADE_DEFAULT_DELAY_DESC describes the upgrade default delay
 	UPGRADE_DEFAULT_DELAY_DESC = "Cluster upgrade to version %s is experiencing a delay whilst it performs necessary pre-upgrade procedures. The upgrade will continue to retry. This is an informational notification and no action is required."
@@ -167,7 +165,7 @@ func createFailureDescription(uc *v1alpha1.UpgradeConfig) string {
 
 	switch failedCondition.Type {
 	case v1alpha1.IsClusterUpgradable:
-		description = fmt.Sprintf(UPGRADE_UPGRADEABLE_FAILED_DESC, uc.Spec.Desired.Version)
+		description = failedCondition.Message
 	case v1alpha1.UpgradePreHealthCheck:
 		description = fmt.Sprintf(UPGRADE_PREHEALTHCHECK_FAILED_DESC, uc.Spec.Desired.Version)
 	case v1alpha1.ExtDepAvailabilityCheck:
--- a/pkg/upgraders/upgradeable.go
+++ b/pkg/upgraders/upgradeable.go
@@ -8,16 +8,15 @@ import (
 	"github.com/go-logr/logr"
 	configv1 "github.com/openshift/api/config/v1"
 	upgradev1alpha1 "github.com/openshift/managed-upgrade-operator/pkg/apis/upgrade/v1alpha1"
-	cv "github.com/openshift/managed-upgrade-operator/pkg/clusterversion"
 )
 
 func (c *clusterUpgrader) IsUpgradeableVersion(ctx context.Context, logger logr.Logger) (bool, error) {
 	v := &configv1.ClusterVersion{}
-	currentVersion, err := cv.GetCurrentVersion(v)
+	currentVersion, err := c.cvClient.GetClusterVersion()
 	if err != nil {
 		return false, err
 	}
-	parsedCurrentVersion, err := semver.Parse(currentVersion)
+	parsedCurrentVersion, err := semver.Parse(currentVersion.Status.History[0].Version)
 	if err != nil {
 		return false, err
 	}
--- a/pkg/upgraders/upgradeable.go
+++ b/pkg/upgraders/upgradeable.go
@@ -3,7 +3,6 @@ package upgraders
 import (
 	"context"
 	"fmt"
-	"strings"
 
 	"github.com/blang/semver"
 	"github.com/go-logr/logr"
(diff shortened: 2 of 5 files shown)


pkg/upgraders/upgradeable.go
```diff
@@ -0,0 +1,74 @@
...
+	cV := &configv1.ClusterVersion{}
+	uC := &upgradev1alpha1.UpgradeConfig{}
+
+	//Check if the cluster is already upgrading -
+	upgradeCommenced, err := c.cvClient.HasUpgradeCommenced(c.upgradeConfig)
+	if err != nil {
+		return false, err
+	}
+	if upgradeCommenced {
+		logger.Info(fmt.Sprintf("Skipping upgrade step %s", upgradev1alpha1.UpgradePreHealthCheck))
```
mrbarge, line 24:
```suggestion
		logger.Info(fmt.Sprintf("Skipping upgrade step %s", upgradev1alpha1.IsClusterUpgradable))
```

pkg/upgraders/upgradeable.go
```diff
@@ -0,0 +1,74 @@
...
+		logger.Error(err, fmt.Sprintf("Failed to parse current cluster version %s as semver", cvVersion))
+		return false, err
+	}
+
+	// get the desired clusterversion from upgradeconfig
+	ucVersion := uC.Spec.Desired.Version
+	if err != nil {
+		logger.Error(err, "failed to get current upgrade config desired version")
+		return false, err
+	}
```
mrbarge, line 46: There is no need to check an err here.

pkg/upgraders/upgradeable.go
```diff
@@ -0,0 +1,74 @@
...
+		logger.Error(err, fmt.Sprintf("failed to parse upgrade config desired version %s as semver", ucVersion))
+		return false, err
+	}
+
+	// check if the clusterversion has "Upgradeable" else return true
+	for _, val := range cV.Status.Conditions {
+		if val.Type != configv1.OperatorUpgradeable {
+			return true, nil
+		}
+	}
```
mrbarge, line 59: In its current form, this is going to return that the cluster is upgradeable on the very first condition it sees which has a type which is not "upgradeable". 

What we want to happen here is to check if there is no Upgradeable condition type at all. If that is the case, then we're not on a version of OCP where we can even look at Upgradeable / ack gates, so we should default to returning true.

So you'll want something like this (note, I have not tested this code):
```suggestion
	// check if the clusterversion has "Upgradeable" else return true
	hasUpgradeableCondition := false
	for _, val := range cV.Status.Conditions {
		if val.Type == configv1.OperatorUpgradeable {
			hasUpgradeableCondition = true
			break
		}
	}
	if !hasUpgradeableCondition {
	    // CVO has no Upgradeable condition to gate upgrades, so we can infer the cluster is upgradeable
	    return true, nil
	}
```

pkg/upgraders/upgradeable.go
```diff
@@ -0,0 +1,74 @@
...
+
+	for _, condition := range cV.Status.Conditions {
+		if condition.Type == configv1.OperatorUpgradeable && condition.Status == configv1.ConditionTrue {
+			return true, nil
+		} else if condition.Type == configv1.OperatorUpgradeable && condition.Status == configv1.ConditionFalse {
```
mrbarge, line 64: Personal stylistic preference, but to me code looks and feels simpler without elseifs.

```suggestion
		if condition.Type == configv1.OperatorUpgradeable && condition.Status == configv1.ConditionFalse {
```

mrbarge:
cc @Tessg22 FYI, above is my initial review on the code so far

pkg/upgraders/upgradeable.go
```diff
@@ -0,0 +1,74 @@
...
+			// if the upgradeable is false then we need to check the current version with upgrade version for y-stream update
+			if parsedUcVersion.Major >= parsedCvVersion.Major && parsedUcVersion.Minor > parsedCvVersion.Minor {
+				return false, err
+			} else if parsedUcVersion.Major >= parsedCvVersion.Major && parsedUcVersion.Minor == parsedCvVersion.Minor {
+				return true, nil
+			}
+		}
+	}
+	return false, nil
```
mrbarge, line 73: See my earlier note. We could just make this return `true` by default and not worry about the earlier check to see whether `Upgradeable` exists as a condition type at all.

mrbarge:
Leaving one more comment on this PR. I think it will be relatively straightforward to get specific service log messaging going that can inform the cluster owner that the upgrade failed specifically due to the upgrade gate issue.

These changes will mainly need to occur in the event manager code (https://github.com/openshift/managed-upgrade-operator/blob/master/pkg/eventmanager/eventmanager.go).

Specifically:

- A new description const for the 'failed' event states. (a 'delayed' one is not necessary IMO)

```
	// UPGRADE_UPGRADEABLE_FAILED_DESC describes the upgradeable check failure
	UPGRADE_UPGRADEABLE_FAILED_DESC = "Cluster upgrade to version %s was cancelled because manual acknowledgement is required concerning the removal of deprecated APIs. Please review the Knowledge Base article at https://access.redhat.com/articles/6329921 and perform the acknowledgement procedure if you are not using workloads that require the deprecated APIs and wish to proceed with the upgrade. Automated upgrades will be retried on their next scheduling cycle. If you have manually scheduled an upgrade instead, it must be rescheduled."
```

- Update the `createFailureDescription` function, in particular [this switch condition](https://github.com/openshift/managed-upgrade-operator/blob/24610af860367e90a79c76a1543a37466dea5f76/pkg/eventmanager/eventmanager.go#L166), to use the aforementioned `UPGRADE_UPGRADEABLE_FAILED_DESC` const if the failedCondition is `IsClusterUpgradable`.

pkg/upgraders/upgradeable.go
```diff
@@ -0,0 +1,65 @@
...
+	"context"
+	"fmt"
+
+	"github.com/blang/semver"
+	"github.com/go-logr/logr"
+	configv1 "github.com/openshift/api/config/v1"
+	upgradev1alpha1 "github.com/openshift/managed-upgrade-operator/pkg/apis/upgrade/v1alpha1"
+)
+
+func (c *clusterUpgrader) IsUpgradeableVersion(ctx context.Context, logger logr.Logger) (bool, error) {
```
mrbarge, line 13: This doesn't need to be an exported function.

pkg/upgraders/upgradeable.go
```diff
@@ -0,0 +1,65 @@
...
+	v := &configv1.ClusterVersion{}
+	currentVersion, err := c.cvClient.GetClusterVersion()
```
mrbarge, line 15: You can use cvClient's `GetCurrentVersion()` here instead, which I'd recommend. You can use this to pass directly into the `semver.Parse()` call below.

pkg/upgraders/upgradeable.go
```diff
@@ -0,0 +1,65 @@
...
+	if err != nil {
+		return false, err
+	}
+	parsedCurrentVersion, err := semver.Parse(currentVersion.Status.History[0].Version)
```
mrbarge, line 19: Can't always guarantee that the first element of CV's history is going to be the current version. For that reason, I recommend using the helper `GetCurrentVersion()` function instead and use it.

pkg/upgraders/upgradeable.go
Tessg22: When I was looking at `GetCurrentVersion()` function, it takes as input the value from `GetClusterVersion()` thus I kept both. https://github.com/openshift/managed-upgrade-operator/blob/master/pkg/clusterversion/cv.go#L205. Please let me know, if it's correct in the latest commit.

mrbarge: That looks great, thank you!

mrbarge:
One final comment in terms of testing. First off, _thanks_ for adding tests, that's awesome!

If they could be expanded to cover some additional scenarios that would be great and is the last thing this PR really needs.

- If the cluster is performing a z-stream upgrade: It should perform upgrade
- If the cluster is performing a y-stream upgrade:
  - If the cluster has no Upgradeable condition: It should perform upgrade
  - If the cluster is Upgradeable: it should perform upgradeable
  - If the cluster is not Upgradeable: it should not perform upgrade  (it has this test already)

pkg/eventmanager/eventmanager.go
```diff
@@ -20,6 +20,8 @@ const (
...
 	// UPGRADE_SCALE_FAILED_DESC describes the upgrade scaling failed
 	UPGRADE_SCALE_FAILED_DESC = "Cluster upgrade to version %s was cancelled during the Scale-Up Worker Node step. A temporary additional worker node was unable to be created to temporarily house workloads, so the upgrade did not proceed. Automated upgrades will be retried on their next scheduling cycle. If you have manually scheduled an upgrade instead, it must now be rescheduled."
+	// UPGRADE_UPGRADEABLE_FAILED_DESC describes the upgradeable check failure
+	UPGRADE_UPGRADEABLE_FAILED_DESC = "Cluster upgrade to version %s was cancelled because manual acknowledgement is required concerning the removal of deprecated APIs. Please review the Knowledge Base article at https://access.redhat.com/articles/6329921 and perform the acknowledgement procedure if you are not using workloads that require the deprecated APIs and wish to proceed with the upgrade. Automated upgrades will be retried on their next scheduling cycle. If you have manually scheduled an upgrade instead, it must be rescheduled."
```
jharrington22, line 24: @Tessg22 @rafael-azevedo the admin ack gate isn't specifically designed to handle API just depreciation. In future releases there will be other related checks. Is there general documentation on these gates in public OCP docs? This doc is specific to 4.9

rafael-azevedo: I don't believe there is general documentation as far as I am aware so far.

mrbarge: This is a fair point. I don't believe, or at least haven't seen, general documentation around the gates that isn't just the enhancements proposal.

But to solve this specific problem, I think the way to go might be to extract the `Reason` from the `clusterversion` Upgradeable condition and present that as the message here.

For example, for the specific 4.9 issue, the message is:
```
Kubernetes 1.22 and therefore OpenShift 4.9 remove several APIs which require admin consideration. Please see
        the knowledge article https://access.redhat.com/articles/6329921 for details and instructions.
```

So I would be in favour of the eventmanager dynamically generating the failure message based on some wrapper text + the CVO message. 

```
Cluster upgrade to version %s was cancelled because manual acknowledgement of upcoming release changes is required. <<REASON HERE>>. Please perform the acknowledgement procedure if you wish to proceed with an upgrade, then re-schedule your upgrade.  Automated upgrades will be retried on their next scheduling cycle. If you have manually scheduled an upgrade instead, it must be rescheduled.
```

The approach of the `createDelayedDescription()` function, which dynamically generates a delayed message based on the conditions causing the delay, could be modelled upon here. Similar deal, just getting stuff out of CVO instead (use the `cv` package to help)

If you want to give this a go, let me know, otherwise we can add it to the backlog to do before 4.10.

Tessg22: @mrbarge did you want to say `createFailureDescription()` instead of `createDelayedDescription()`? We could make use of `failedCondition` var that is already defined at [line 157](https://github.com/openshift/managed-upgrade-operator/blob/edb5b0f9e13109478b22dac3163e8fe3c5edaf16/pkg/eventmanager/eventmanager.go#L157). Then just replace the hardcoded message with `description = failedCondition.Message`. I think there is no need to create a separate card for this. I will test if there are any hidden complications or edge cases.

Tessg22: I am wondering if there will be more than one check/condition in the newer versions.

rafael-azevedo: @Tessg22 In the future it is a possibility there could be more then 1 ack condition.

wking: It's also broader than just acks.  For example, you could be `Upgradeable=False` in the ClusterVersion because you have excessive node skew: openshift/cluster-kube-apiserver-operator#1199.  I'd recommend just passing along any `Upgradeable=False` `reason` and `message`, with wrapping wording drawn from [the API docs][1] if you want to provide more context on the condition (although I'd hope the `message` would speak for itself, and think it's bug-worthy if it does not).

[1]: https://github.com/openshift/api/blob/cc0db1116639638254e87a564902833f1ee006d5/config/v1/types_cluster_operator.go#L179-L189

dofinn: +1 @wking

Tessg22: @wking, do you know how multiple "Upgradeable=False" conditions will be displayed in CVO (if at all)?

Tessg22: > I'd recommend just passing along any Upgradeable=False reason and message with wrapping wording

What do you think about this wording?
`fmt.Errorf("Cluster upgrade to version %s is canceled with the reason of %s containing message that %s Automated upgrades will be retried on their next scheduling cycle. If you have manually scheduled an upgrade instead, it must be rescheduled.", desiredVersion, condition.Reason, condition.Message)
`

Example: Cluster upgrade to version 1.2.3 is canceled with the reason of IsClusterUpgradable not done containing message that Kubernetes 1.22 and therefore OpenShift 4.9 remove several APIs which require admin consideration. Please see the knowledge article https://access.redhat.com/articles/6329921 for details and instructions. Automated upgrades will be retried on their next scheduling cycle. If you have manually scheduled an upgrade instead, it must be rescheduled.

mrbarge: @Tessg22 @anjoshi24 I had a look at the CVO code and yes CVO will [list multiple reasons](https://github.com/openshift/cluster-version-operator/blob/06ec265e3a3bf47b599e56aec038022edbe8b5bb/pkg/cvo/upgradeable.go#L78) if they exist.

Given that, you should be able to just take the `Message` from the CVO condition and use that as the reliable messaging of why the cluster was not upgradeable.

Tessg22: In latest commit, I have updated the wording of error message and input of IsClusterUpgradable notification.

mrbarge: @Tessg22 That looks good, thank you.

PR #298, opened 2022-02-15
Fix_machinesetscaler_bug_for_OSD-6692

anjoshi24:
Signed-off-by: Aniket Joshi <anjoshi@redhat.com>

### What type of PR is this?
Bug

Fixing the machinesetscaler bug to reduce MUO confusion of additional machines during upgrades.

OSD-6692


pkg/scaler/machineSetScaler.go
```diff
@@ -200,6 +200,8 @@ func (s *machineSetScaler) EnsureScaleDownNodes(c client.Client, nds drain.NodeD
 	err = c.List(context.TODO(), originalMachineSets, []client.ListOption{
 		client.InNamespace(MACHINE_API_NAMESPACE),
 		NotMatchingLabels{LABEL_UPGRADE: "true"},
+		client.MatchingLabels{"hive.openshift.io/machine-pool": "worker"},
+		client.MatchingLabels{"hive.openshift.io/machine-pool": "infra"},
```
mrbarge, line 204: I'm not really sure how this would address the problem, having multiple label selectors match the same label in the `List` is only going to end up with one of them succeeding, not both.

ie. what this call is doing is equivalent to
`oc get machinesets -n openshift-machine-api -l hive.openshift.io/machine-pool=worker -l hive.openshift.io/machine-pool=infra`
which is only going to return the infra machineset as it was the last label selector in the list (reverse them, and it'll be the worker returned)

mrbarge:
A couple of notes here.

First, please rebase your fork off master; it is quite behind at the moment (pre boilerplate onboarding). It is not going to be possible to reasonably test this until it is rebased.

Second is that I don't think the change this PR is proposing is going to address the problem.

The underlying issue is that if a machine is created, which is not a member of any machine set, it confuses the Scaler when scaling down.

This is because the scaler's current logic in deciding "have we scaled down is":

- Get a count of all non-master nodes in the cluster
- Get a count of all machines belonging to machinesets in the cluster that aren't upgrade-capacity-scaled machinesets.
- Do those counts equal? If so, it's scaled down. If not, it's not scaled down. 

The problem is, however, that these counts will never become equal if there is a machine that's not part of a machineset.

The ultimate question I think we want the 'scale down' code to solve is:

- There must not be any machines left on the cluster which originally spawned from upgrade-scaled machinesets.

I _think_ the solution to this should be as simple as getting a list of all machines which have the upgrade label set to `true` on them. If that list is empty then we can consider that the scale down is complete. I don't see a reason to complicate matters by bringing nodes into the picture. Maybe I'm forgetting something, because it seems like something obvious we would have done to begin with.

pkg/scaler/machineSetScaler.go
```diff
@@ -108,33 +107,18 @@ func (s *machineSetScaler) EnsureScaleDownNodes(c client.Client, nds drain.NodeD
...
 		return false, err
 	}
 
-	var desiredNodeCount int32 = 0
-	for _, ms := range originalMachineSets.Items {
-		desiredNodeCount += *ms.Spec.Replicas
-	}
-
-	if desiredNodeCount != int32(len(nonMasterNodes.Items)) {
+	if len(originalMachines.Items) != 0 {
```
mrbarge, line 121: It would be helpful if it logged how many upgrade machines are still waiting to be terminated, if there is more than 0 waiting to be terminated.

pkg/scaler/scaling_test.go
```diff
@@ -439,39 +439,21 @@ var _ = Describe("Node scaling tests", func() {
...
-						},
 					},
 				},
 			}
 			gomock.InOrder(
 				mockKubeClient.EXPECT().List(gomock.Any(), gomock.Any(), []client.ListOption{
 					client.InNamespace(MACHINE_API_NAMESPACE),
 					client.MatchingLabels{LABEL_UPGRADE: "true"},
-				}).SetArg(1, *upgradeMachinesets),
+				}).AnyTimes().SetArg(1, *upgradeMachinesets),
```
mrbarge, line 456: Not sure about the reason for this change? If anything, please make it `Times(1)`

pkg/scaler/scaling_test.go
```diff
@@ -561,17 +512,13 @@ var _ = Describe("Node scaling tests", func() {
 						Expect(found).To(BeTrue())
 						return nil
 					}).Times(2),
-				mockKubeClient.EXPECT().List(gomock.Any(), gomock.Any()).SetArg(1, *nodes),
 				mockKubeClient.EXPECT().List(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).SetArg(1, *upgradeMachines),
```
mrbarge, line 515: This should not be removed, nor should the nodes variables declared above it be removed. I think there is a misunderstanding happening of what the `gomock.InOrder` function is performing. 

The list of calls happening in `gomock.InOrder` are mocking calls that the function would otherwise make to the kube API. When running these test cases there is no kube API for the code to call, so the tests mock it out and act like a fake client that returns responses. 

There is a big block of code in `ensureScaleDownNodes` which runs _before_ the change that this PR is making, which is involved around retrieving nodes and executing drain strategies on them. None of that code is changing, so the mock calls still need to be there. 

For instance, this line still gets called:
https://github.com/anjoshi24/managed-upgrade-operator/blob/4b27f7710a75ebee0e88b35fb39a5c1e5fb00825/pkg/scaler/machineSetScaler.go#L100
which results in a `List` of Nodes. It is still necessary to have the test set up a mock for this call.

pkg/scaler/scaling_test.go
```diff
@@ -561,17 +512,13 @@ var _ = Describe("Node scaling tests", func() {
...
 				mockDrainStrategy.EXPECT().Execute(gomock.Any()).Return([]*drain.DrainStrategyResult{}, nil),
 				mockDrainStrategy.EXPECT().HasFailed(gomock.Any()).Return(false, nil),
 				mockKubeClient.EXPECT().List(gomock.Any(), gomock.Any(), []client.ListOption{
 					client.InNamespace(MACHINE_API_NAMESPACE),
 					NotMatchingLabels{LABEL_UPGRADE: "true"},
-				}).SetArg(1, *originalMachineSets),
-				mockKubeClient.EXPECT().List(gomock.Any(), gomock.Any(), []client.ListOption{
-					NotMatchingLabels{machinery.MasterLabel: ""},
-				}).SetArg(1, *masterNodes),
+				}).SetArg(1, *originalMachines),
```
mrbarge, line 521: The purpose of this test is/was to verify the path of `EnsureScaleDownNodes` that returns `true`. In essence:

- that the function executes a drain strategy 
- that no upgrade-related machines remain in the cluster

But this mock is returning a machine list with 1 in it. Which is not going to get a 'everything is scaled down, return true' path. So really this should be returning an empty machine list.

That probably highlights that we're missing a test in this suite - we want a test exactly like this, but simulating the scenario where the drain strategies have been executed but an upgrade machine is still being returned. And in that scenario the test should expect that `EnsureScaleDownNodes` returns `false`.

anjoshi24: I have fixed the current block with it expected to return an empty machine list. 

Also added another test in the suite where it will target the scenario where the drain strategies have been executed but an upgrade machine is still being returned and then expects `EnsureScaleDownNodes` returns `false`.

pkg/scaler/scaling_test.go
anjoshi24: Added the required list call for nodes.

pkg/scaler/scaling_test.go
anjoshi24: Fixed and tests passed.

pkg/scaler/machineSetScaler.go
anjoshi24: Added a block to print the machines from machine list using the logger.

mrbarge:
/lgtm
Please unhold when you are ready to verify this in staging/e2e.

PR #303, opened 2022-03-18
updated clustermaps

jwai7:
### What type of PR is this?
Cleanup

Added missing configManager properties to clustermap
Added test prefix to names in rolebindings


mrbarge:
Thanks for the changes! Can I ask that you please submit this from a separate branch in your fork rather than committing straight onto master? This will be an essential habit to get into, so it'll be good to use this PR as an excuse to familiarise with the process.

PR #309, opened 2022-04-11
[OSD-10140] latest tag is being removed from the quay repo, use specific version for managed-upgrade-operator image

T0MASD:
latest tag is being removed from the quay repo, use specific version for managed-upgrade-operator image

### What type of PR is this?
_(refactor)_

_Fixes #_ [OSD-10140](https://issues.redhat.com//browse/OSD-10140)

- [x] Ran make test

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -40,7 +40,7 @@ spec:
         - name: managed-upgrade-operator
           # Replace this with the built image name
           # This will get replaced on deploy by /hack/generate-operator-bundle.py
-          image: quay.io/app-sre/managed-upgrade-operator:v0.1.809-46d986a
+          image: quay.io/app-sre/managed-upgrade-operator:GENERATED
           command:
           - managed-upgrade-operator
           imagePullPolicy: Always
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -40,7 +40,7 @@ spec:
         - name: managed-upgrade-operator
           # Replace this with the built image name
           # This will get replaced on deploy by /hack/generate-operator-bundle.py
-          image: quay.io/app-sre/managed-upgrade-operator:latest
+          image: quay.io/app-sre/managed-upgrade-operator:v0.1.809-46d986a
           command:
           - managed-upgrade-operator
           imagePullPolicy: Always
--- a/test/deploy/operator.yaml
+++ b/test/deploy/operator.yaml
@@ -18,7 +18,7 @@ spec:
         - name: managed-upgrade-operator
           # Replace this with the built image name
           # This will get replaced on deploy by /hack/generate-operator-bundle.py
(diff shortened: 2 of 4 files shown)


deploy/operator.yaml
```diff
@@ -40,7 +40,7 @@ spec:
         - name: managed-upgrade-operator
           # Replace this with the built image name
           # This will get replaced on deploy by /hack/generate-operator-bundle.py
-          image: quay.io/app-sre/managed-upgrade-operator:latest
+          image: quay.io/app-sre/managed-upgrade-operator:v0.1.809-46d986a
```
mrbarge, line 43: (this comment applies to this line as well as the one in `csv-template.yaml`)

I think one of the concerns that led to the original ticket was that having a valid version set here means that there is the possibility that - if some pipelines happen to not replace the image correctly and also didn't fail - MUO could be deployed running an old version and no-one would notice unless they looked at it.

So setting it to a real version still leaves that risk in place. It should be a invalid version such that it _should_ cause errors if somehow this deployment were go out, which would highlight a failure in the deployment pipeline processes.

test/deploy/operator.yaml
```diff
@@ -18,7 +18,7 @@ spec:
         - name: managed-upgrade-operator
           # Replace this with the built image name
           # This will get replaced on deploy by /hack/generate-operator-bundle.py
-          image: quay.io/app-sre/managed-upgrade-operator:latest
+          image: quay.io/app-sre/managed-upgrade-operator:v0.1.809-46d986a
```
mrbarge, line 21: Similarly I feel like this benefits from being an invalid tag because it forces the SRE testing the operator to think about "what version do I _want_ to test?", rather than not thinking about that at all and having some old version be used.

deploy/operator.yaml
mrbarge: I've had a go at pushing this change to the branch.

test/deploy/operator.yaml
mrbarge: I've had a go at pushing this change to the branch.

rendhalver:
Should we be updating the docs as well?
They currently say to use `latest` as the tag to push.
https://github.com/openshift/managed-upgrade-operator/blob/fd0a2dc5cb1287b1e3cbd55a205497478ec5bdf1/docs/development.md#L216

mrbarge:
> Should we be updating the docs as well? They currently say to use `latest` as the tag to push.
> 
> https://github.com/openshift/managed-upgrade-operator/blob/fd0a2dc5cb1287b1e3cbd55a205497478ec5bdf1/docs/development.md#L216

In this case I think it's fine, when testing in a personal repo it's okay to use the `latest` tag (or really any tag they want). They'll still need to update the `image` in the file either way. I've pushed a small update to the docs to clarify that process though.

PR #312, opened 2022-04-28
OSD-11105 updated error message in case the cluster cannot be upgraded

Nikokolas3270:
Distinguishing ROSA cluster from OSD cluster using infrastructure resource

### What type of PR is this?
_bug fix_

This PR align the code with the following spec:
https://docs.google.com/spreadsheets/d/1Ghx1J_hjqPIzEPmVuIkmTDW_HMFF2-tw0BO6bQ-6ecI/edit#gid=0&range=14:15
Remark that the error message is no longer the same regarding the cluster is a ROSA cluster or a normal OSD cluster.

[OSD-11105](https://issues.redhat.com//browse/OSD-11105)

`ROSA` vs `OSD` cluster type is retrieved through the `infrastructure` object. On an openshift cluster you can view it running `oc get infrastructure -oyaml`.

### DIFF
--- a/pkg/upgraders/upgradeable.go
+++ b/pkg/upgraders/upgradeable.go
@@ -3,13 +3,40 @@ package upgraders
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/blang/semver"
 	"github.com/go-logr/logr"
 	configv1 "github.com/openshift/api/config/v1"
 	upgradev1alpha1 "github.com/openshift/managed-upgrade-operator/pkg/apis/upgrade/v1alpha1"
 	cv "github.com/openshift/managed-upgrade-operator/pkg/clusterversion"
+
+	"k8s.io/apimachinery/pkg/types"
 )
+
+const (
+	upgradeCancelledArticleNumberForOSD = 6657541
+	upgradeCancelledArticleNumberForROSA = 6541901
+)
+
+func (c *clusterUpgrader) isRosaCluster(ctx context.Context) (bool, error) {
+	infraConfig := &configv1.Infrastructure{}
+	if err := c.client.Get(ctx, types.NamespacedName{Name: "cluster"}, infraConfig); err != nil {
+		return false, fmt.Errorf("failed fetching infrastructure config: %w", err)
+	}
+
+	platformStatus := infraConfig.Status.PlatformStatus
+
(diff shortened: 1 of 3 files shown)


pkg/upgraders/upgradeable_test.go
```diff
@@ -108,61 +110,73 @@ var _ = Describe("UpgradableCheckStep", func() {
...
+
+					var kbArticleNumber int
+
+					if isRosaCluster {
+						kbArticleNumber = 6541901
+					} else {
+						kbArticleNumber = 6657541
+					}
+
+					Expect(err.Error()).Should(MatchRegexp(fmt.Sprintf("Cluster upgrade maintenance to version .* has been cancelled due to unacknowledged user actions. See https://access.redhat.com/solutions/%d for more details.", kbArticleNumber)))
```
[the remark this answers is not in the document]
Nikokolas3270: Then the test will fail which is what we want.
Here we are checking that the upgrade is not possible for a very specific reason :)

mrbarge:
This will need additions to MUO's [ClusterRole](https://github.com/openshift/managed-upgrade-operator/blob/master/deploy/cluster_role.yaml) to allow for additional RBAC required to get/list `infrastructures` - can that please be added with this PR?

pkg/upgraders/upgradeable.go
```diff
@@ -41,9 +63,33 @@ func (c *clusterUpgrader) IsUpgradeable(ctx context.Context, logger logr.Logger)
...
+		return false, err
+	}
+
+	var kbArticleNumber int
+
+	if isRosaCluster {
+		kbArticleNumber = 6541901
+	} else {
+		kbArticleNumber = 6657541
+	}
```
mrbarge, line 77: Can these numbers please be turned into const declarations outside of the function?

Nikokolas3270: Sure, will turn them into `const`

Nikokolas3270:
> This will need additions to MUO's [ClusterRole](https://github.com/openshift/managed-upgrade-operator/blob/master/deploy/cluster_role.yaml) to allow for additional RBAC required to get/list `infrastructures` - can that please be added with this PR?

Definitely, thanks for pointing that out :)

PR #314, opened 2022-05-10
OSD-10033; OSD-10032 - Add metric to track paging alerts

tnierman:
### What type of PR is this?
**Feature**

Fixes [OSD-10032](https://issues.redhat.com/browse/OSD-10032) and [OSD-10033](https://issues.redhat.com/browse/OSD-10033)

This PR adds additional metrics for [OSD-10033](https://issues.redhat.com//browse/OSD-10033) and [OSD-10032](https://issues.redhat.com//browse/OSD-10032) to track alerts that paged SREP Primary during the latest upgrade. It does this by running additional logic during the `upgradeconfig` reconcile loop at the terminal stages of the upgrade (either when the upgrade is complete or has failed). This logic triggers a lookup via Prometheus query to determine which (if any) `pagingAlerts` fired during the upgrade, and adds them to the new `upgradeoperator_upgrade_alerted` metric. This new metric will be exported to Observatorium-MST ([relevant PR](https://github.com/openshift/managed-cluster-config/pull/1164)) for long-term storage and later analysis. 

- The new metric `upgradeoperator_upgrade_alerted` added in this Pull Request should not be cleared on upgrade completion, whereas existing metrics should. Because of this, this PR divides the metrics into `ephemeral` and `persistent` metrics - those that should and should not be reset on upgrade completion, respectively. Additionally, the function `ResetAllMetrics` has been renamed to `ResetEphemeralMetrics` for clarity. 

- While the new metric shouldn't be cleared, only updated with each upgrade completion, the alert *is* cleared when the `managed-upgrade-operator` pod is restarted. I haven't found an elegant way to recover this metric between pod restarts; options I've considered include writing the results to a configmap, re-querying prometheus on operator start-up, or reconciling against another object (`clusterversion` objects have upgrade timestamps which do not appear to match the ones in `upgradeconfigs`, but maybe there's another object that I haven't considered?). Currently, I think accounting for this would make sense as a follow-up task, but I understand if this is something that we want to resolve from the get-go as well. 

### DIFF
--- a/pkg/controller/upgradeconfig/upgradeconfig_controller.go
+++ b/pkg/controller/upgradeconfig/upgradeconfig_controller.go
@@ -130,8 +130,8 @@ func (r *ReconcileUpgradeConfig) Reconcile(ctx context.Context, request reconcil
 			// Request object not found, could have been deleted after reconcile request.
 			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
 			// Return and don't requeue
-			metricsClient.ResetAllMetrics()
-			reqLogger.Info("Reset all the metrics due to no upgrade config present.")
+			metricsClient.ResetEphemeralMetrics()
+			reqLogger.Info("Reset metrics due to no upgrade config present.")
 			return reconcile.Result{}, nil
 		}
 		// Error reading the object - requeue the request.
@@ -272,7 +272,8 @@ func (r *ReconcileUpgradeConfig) Reconcile(ctx context.Context, request reconcil
 		return r.upgradeCluster(upgrader, instance, reqLogger)
 	case upgradev1alpha1.UpgradePhaseUpgraded:
 		reqLogger.Info("Cluster is already upgraded")
-		return reconcile.Result{}, nil
+		err = reportUpgradeMetrics(metricsClient, instance.Name, instance.Spec.Desired.Version, history.StartTime.Time, history.CompleteTime.Time)
+		return reconcile.Result{}, err
 	case upgradev1alpha1.UpgradePhaseFailed:
 		reqLogger.Info("Cluster has failed to upgrade")
 		return reconcile.Result{}, nil
@@ -299,6 +300,17 @@ func (r *ReconcileUpgradeConfig) upgradeCluster(upgrader cub.ClusterUpgrader, uc
 	me = multierror.Append(err, me)
 
 	return reconcile.Result{RequeueAfter: 1 * time.Minute}, me.ErrorOrNil()
+}
+
+// reportUpgradeMetrics updates prometheus with statistics from the latest upgrade
+func reportUpgradeMetrics(metricsClient metrics.Metrics, name string, version string, upgradeStart time.Time, upgradeEnd time.Time) error {
+	upgradeAlerts, err := metricsClient.AlertsFromUpgrade(upgradeStart, upgradeEnd)
+	if err != nil {
+		return err
+	}
+
+	metricsClient.UpdateMetricUpgradeResult(name, version, upgradeAlerts)
+	return nil
 }
 
 // ManagedUpgradePredicate is used for managing predicates of the UpgradeConfig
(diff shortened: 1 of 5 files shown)


pkg/controller/upgradeconfig/upgradeconfig_controller.go
```diff
@@ -272,10 +272,12 @@ func (r *ReconcileUpgradeConfig) Reconcile(ctx context.Context, request reconcil
 		return r.upgradeCluster(upgrader, instance, reqLogger)
 	case upgradev1alpha1.UpgradePhaseUpgraded:
 		reqLogger.Info("Cluster is already upgraded")
-		return reconcile.Result{}, nil
+		err = reportUpgradeMetrics(metricsClient, instance.Name, instance.Spec.Desired.Version, history.StartTime.Time, history.CompleteTime.Time)
+		return reconcile.Result{}, err
 	case upgradev1alpha1.UpgradePhaseFailed:
 		reqLogger.Info("Cluster has failed to upgrade")
-		return reconcile.Result{}, nil
+		err = reportUpgradeMetrics(metricsClient, instance.Name, instance.Spec.Desired.Version, history.StartTime.Time, history.CompleteTime.Time)
```
mrbarge, line 279: May be worth doublechecking this, but in the event of an upgradefailure I don't believe `CompleteTime` is set (I see only one instance that writes to in the code, and that's when the phase is `Upgraded`), and I'm not sure how that might throw off the `reportUpgradeMetrics` function.

mrbarge:
On the whole, seems like it could/should work. Left a small comment that might need looking at.

I must highlight @ravitri put in a lot of thoughtful investigation into thinking about this topic recently (I apologise, I was not aware there were multiple teams looking at this same thing!) in OSD-6997. The document (linked in that ticket) weighed up the pros and cons of whether this could be achieved simply via Telemeter without needing to introduce a new metric, given that Telemetry would have a record of alerts firing, vs the need to introduce a metric on-cluster. 

I encourage a read of that document as it does provide a lot of food for thought around the topic of making use of this sort of data for SLOs.

tnierman:
> On the whole, seems like it could/should work. Left a small comment that might need looking at.
> 
> I must highlight @ravitri put in a lot of thoughtful investigation into thinking about this topic recently (I apologise, I was not aware there were multiple teams looking at this same thing!) in OSD-6997. The document (linked in that ticket) weighed up the pros and cons of whether this could be achieved simply via Telemeter without needing to introduce a new metric, given that Telemetry would have a record of alerts firing, vs the need to introduce a metric on-cluster.
> 
> I encourage a read of that document as it does provide a lot of food for thought around the topic of making use of this sort of data for SLOs.

Thanks for pointing this out, I was not aware work had already been done on this front either! The "MUO SLO Design document" does provide some great analysis, and I see that there was a bit of discussion regarding the use of Telemetry vs additional metrics there as well, which I've added to. 

I'm not actually all that familiar with Telemetry, but based on the document linked in the title of the [SRE_dashboards_metrics](https://github.com/openshift/ops-sop/blob/master/v4/setup/SRE_dashboards_metrics.asciidoc) SOP, it seems as though the Cluster Monitoring Operator exports a set list of metrics defined in [this ConfigMap](https://github.com/openshift/cluster-monitoring-operator/blob/master/manifests/0000_50_cluster-monitoring-operator_04-config.yaml)? Do we have a way to add additional metrics to be exported? 

I'm certainly not opposed to closing this PR in favor of a Telemetry-based approach to SLI-tracking, if it's deemed the better solution.

mrbarge:
> I'm certainly not opposed to closing this PR in favor of a Telemetry-based approach to SLI-tracking, if it's deemed the better solution.

Oh, I didn't mean to imply that with my reply, I'm sorry. :)  I think this PR is fine, ultimately I think Ravi's document lent towards us doing this on-cluster rather than Telemetry and I've not thought through doing it on the Telemetry end too seriously.

tnierman:
No worries, just want to make sure the solution used is well-thought out! 

In addition to removing metrics when the upgrade failed, I've also renamed the metric used here to `upgradeoperator_upgrade_result`. This way, the metric represents a more general status of the cluster's recent upgrade(s), which we can build on in the future if need be (Ravi's analysis mentioned that `upgradeStart` and `upgradeComplete` timestamps might be additional metrics we'd like to expose).

In terms of tracking the upgrade results over a 4 week window - I haven't come up with an effective solution yet.

PR #322, opened 2022-06-14
OSD-11002 patch clusterversion instead of update

rendhalver:
### What type of PR is this?
bug

An Update() to clusterversion can result in potentially cause fields to be nullified or emptied in it,
if the client performing the update isn't aware of those fields.
This will impact MUO as it performs an Update to clusterversion when it initiates an upgrade.
Instead of an Update it should perform a Patch.

[OSD-11002](https://issues.redhat.com//browse/OSD-11002)

### DIFF
--- a/pkg/clusterversion/cv.go
+++ b/pkg/clusterversion/cv.go
@@ -275,8 +275,8 @@ func (c *clusterVersionClient) runUpgradeWithImage(cv *configv1.ClusterVersion, 
 
 	if cv.Spec.DesiredUpdate == nil || cv.Spec.DesiredUpdate.Image != desired.Image {
 		logger.Info(fmt.Sprintf("Setting ClusterVersion to Image %s", desired.Image))
-		cv.Spec.DesiredUpdate = &configv1.Update{Image: desired.Image, Version: ""}
-		err := c.client.Update(context.TODO(), cv)
+		patch := []byte(fmt.Sprintf(`{"spec":{"desiredUpdate":{"image": "%s", "version": "" }}}`, desired.Image))
+		err := c.client.Patch(context.TODO(), cv, client.RawPatch(types.StrategicMergePatchType, patch))
 		if err != nil {
 			return false, err
 		}
@@ -289,8 +289,8 @@ func (c *clusterVersionClient) runUpgradeWithChannelVersion(cv *configv1.Cluster
 
 	if cv.Spec.Channel != desired.Channel {
 		logger.Info(fmt.Sprintf("Setting ClusterVersion to Channel %s Version %s", desired.Channel, desired.Version))
-		cv.Spec.Channel = desired.Channel
-		err := c.client.Update(context.TODO(), cv)
+		patch := []byte(fmt.Sprintf(`{"spec":{"Channel": "%s" }}`, desired.Channel))
+		err := c.client.Patch(context.TODO(), cv, client.RawPatch(types.StrategicMergePatchType, patch))
 		if err != nil {
 			return false, err
 		}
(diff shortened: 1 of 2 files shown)


rendhalver:
/hold
Running a few upgrade tests before we get this merged.

pkg/clusterversion/cv.go
```diff
@@ -289,8 +289,8 @@ func (c *clusterVersionClient) runUpgradeWithChannelVersion(cv *configv1.Cluster
 
 	if cv.Spec.Channel != desired.Channel {
 		logger.Info(fmt.Sprintf("Setting ClusterVersion to Channel %s Version %s", desired.Channel, desired.Version))
-		cv.Spec.Channel = desired.Channel
-		err := c.client.Update(context.TODO(), cv)
+		patch := []byte(fmt.Sprintf(`{"spec":{"channel":"%s"}}`, desired.Channel))
```
mrbarge, line 292: As you've encountered during testing, having the hand-written json strings can lead to inadvertent test failures simply due to things such as an extra space character, etc, which is a little bit ugly.

The simplest way would be to switch the patch to using an approach via `StrategicMergeFrom` ([link](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#StrategicMergeFrom)) which generates the patch for you by comparing two objects.

I'd wondered if building up the update using the [struct](https://github.com/openshift/api/blob/master/config/v1/types_cluster_version.go#L364) and then marshalling that to a json string would work, but that would not result in a correct patch (it'd be missing the outer `spec` and `desiredUpdate` layers).

But that would be pretty much re-work of the code and tests to switch to the `StrategicMergeFrom` approach. Maybe you could write it up as a low-hanging-fruit toil maintenance ticket to look at in the future.

pkg/clusterversion/cv.go
```diff
@@ -314,8 +314,8 @@ func (c *clusterVersionClient) runUpgradeWithChannelVersion(cv *configv1.Cluster
 	}
 
 	cv.Spec.Overrides = []configv1.ComponentOverride{}
-	cv.Spec.DesiredUpdate = &configv1.Update{Version: uc.Spec.Desired.Version}
-	err := c.client.Update(context.TODO(), cv)
+	patch := []byte(fmt.Sprintf(`{"spec":{"desiredUpdate":{"version":"%s"}}}`, desired.Version))
```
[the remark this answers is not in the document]
rendhalver: That is an excellent point. I hadn't thought about that.
I will get it updated.

pkg/clusterversion/cv.go
```diff
@@ -275,8 +275,8 @@ func (c *clusterVersionClient) runUpgradeWithImage(cv *configv1.ClusterVersion,
 
 	if cv.Spec.DesiredUpdate == nil || cv.Spec.DesiredUpdate.Image != desired.Image {
 		logger.Info(fmt.Sprintf("Setting ClusterVersion to Image %s", desired.Image))
-		cv.Spec.DesiredUpdate = &configv1.Update{Image: desired.Image, Version: ""}
-		err := c.client.Update(context.TODO(), cv)
+		patch := []byte(fmt.Sprintf(`{"spec":{"desiredUpdate":{"image":"%s","version":null}}}`, desired.Image))
```
bmeng, line 278: not a big deal, but maybe better to change the name here to something like `desiredImage` instead of `patch`?

bmeng: same for below, like desiredChannel, desiredVersion

rendhalver: That's a good idea.
Thanks for the suggestion.

pkg/clusterversion/cv.go
rendhalver: I am trying to work out how to switch to using `MergeFrom` and what needs to change in the tests to get those passing.
I am not sure if I will be finished that today.
I am also wondering if that would be best done in it's own change.

PR #323, opened 2022-06-16
Osd 11612

jwai7:
```
Updated ControlPlaneUpgraded and tests to fallback to upgrade start time if history is not found

### What type of PR is this?
Bug

If the ControlPlaneUpgraded function cannot find the current in-progress update in the clusterVersion history to determine the start time of the upgrade, it will instead use the time that the upgrade was initiated in order to set the UpgradeControlPlaneTimeout metric

Fixes OSD-11612 https://issues.redhat.com/browse/OSD-11612

```

### DIFF
--- a/pkg/upgraders/controlplanestep.go
+++ b/pkg/upgraders/controlplanestep.go
@@ -48,7 +48,7 @@ func (c *clusterUpgrader) ControlPlaneUpgraded(ctx context.Context, logger logr.
 	if history == nil {
 		upgradeStartTime, err = time.Parse(time.RFC3339, c.upgradeConfig.Spec.UpgradeAt)
 		if err != nil {
-			return false, err //handle error later
+			return false, err //error parsing time string
 		}
 	} else {
 		upgradeStartTime = history.StartedTime.Time
--- a/pkg/upgraders/controlplanestep.go
+++ b/pkg/upgraders/controlplanestep.go
@@ -43,14 +43,22 @@ func (c *clusterUpgrader) ControlPlaneUpgraded(ctx context.Context, logger logr.
 
 	isCompleted := c.cvClient.HasUpgradeCompleted(clusterVersion, c.upgradeConfig)
 	history := cv.GetHistory(clusterVersion, c.upgradeConfig.Spec.Desired.Version)
+	var upgradeStartTime time.Time
+	var controlPlaneCompleteTime time.Time
 	if history == nil {
-		return false, err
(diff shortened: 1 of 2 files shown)


pkg/upgraders/controlplanestep_test.go
```diff
@@ -121,6 +121,31 @@ var _ = Describe("ControlPlaneStep", func() {
 			})
 		})
 
+		Context("When that version is not recorded in clusterversion's history", func() {
```
mrbarge, line 124: I would have two tests here, and provide better contexts to describe the test behaviours.

Pseudocode below.

```suggestion
		Context("When that version is not recorded in clusterversion's history", func() {
				BeforeEach(func() {
						// setup clusterVersion here, since it will be common to both tests
				})
				Context("When the upgrade window has expired", func() {
						BeforeEach(func() {
								// setup upgradeconfig here
						})
						// put your existing test here
				})
				Context("When the upgrade window has not yet expired", func() {
						// a new test, that tests to make sure that when the upgrade window has not expired it just returns false, but doesn't set the timeout metric
				})
		})
```

pkg/upgraders/controlplanestep.go
```diff
@@ -43,14 +43,22 @@ func (c *clusterUpgrader) ControlPlaneUpgraded(ctx context.Context, logger logr.
...
+	var upgradeStartTime time.Time
+	var controlPlaneCompleteTime time.Time
 	if history == nil {
-		return false, err
+		upgradeStartTime, err = time.Parse(time.RFC3339, c.upgradeConfig.Spec.UpgradeAt)
+		if err != nil {
+			return false, err //error parsing time string
+		}
+	} else {
+		upgradeStartTime = history.StartedTime.Time
```
mrbarge, line 54: I think you may need to do a check to verify that `history.StartedTime` is not nil / `IsZero` before you go accessing the `Time` in it or else this might cause a panic.

mrbarge: Note that if the StartedTime is nil, we would still want the code logic to be able to set the timeout metric.
ie. if there is never a startedtime and the upgrade window elapses, we still need to set that timeout metric.

mrbarge:
@jwai7 if this is no longer in draft and you feel it is ready for review can you please remove the draft label and add a comment on any testing you've undertaken (and the results).

This helps reviewers know the current status of this PR.

mrbarge:
Thanks!

/approve

Will let another reviewer apply the lgtm.

PR #324, opened 2022-06-27
[OSD-11456] Address gosec lint warnings

a7vicky:
The PR addresses the gosec lint warnings

_Fixes #_[OSD-11456](https://issues.redhat.com/browse/OSD-11456)

### DIFF
--- a/util/access_token.go
+++ b/util/access_token.go
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	pullSecretKey     = ".dockerconfigjson"
+	pullSecretKey     = ".dockerconfigjson" //#nosec G101 -- This is a false positive
 	pullSecretAuthKey = "cloud.openshift.com"
 )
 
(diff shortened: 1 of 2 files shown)


util/access_token.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	pullSecretKey     = ".dockerconfigjson"
+	pullSecretKey     = ".dockerconfigjson" //#nosec G101 -- This is a false positive
```
bmeng, line 16: I am wondering if we should fix the "confusing" name of the const.

And it is interesting there the next line is not being caught?

bmeng:
The fix looks good to me.
One only comment, maybe we should enable the gosec linter as a default one in this repo, as we have the golangci.yml already.

a7vicky:
> The fix looks good to me. One only comment, maybe we should enable the gosec linter as a default one in this repo, as we have the golangci.yml already.

Thanks @bmeng for the review. I have enabled gosec linter in golangci.yml

PR #325, opened 2022-07-01
[OSD-12087] introduce new metric for upgrdeconfig sync timestamp

bmeng:
### What type of PR is this?
_bug/refactor_

Change the logic for the metric which used by UpgradeConfigSyncFailure to avoid some false alerts.

_Fixes #OSD-12087_


mrbarge:
/hold

Why is this PR removing the setting of metrics for failure of the pre-upgrade and post-upgrade health check, or the scaler step?

bmeng:
> /hold
> 
> Why is this PR removing the setting of metrics for failure-correction of the pre-upgrade and post-upgrade health check, or the scaler step?

Just a simple clean up. Those functions are not being used. The current alerts are all based on the failure, where the metric set to 1. And the removed functions are all setting the value to 0.
And we also have the Reset() to set all of them to 0 to avoid the false alert.

PR #333, opened 2022-07-28
OSD-11795 Migrate to Operator-SDK 1.21

ravitri:
### What type of PR is this?
_(refactor)_

This PR is to migrate operator-sdk version from 0.18 to 1.21 version to have recent controller-runtime and kube changes.

_Fixes #_ [OSD-11795](https://issues.redhat.com/browse/OSD-11795)


build/Dockerfile
```diff
@@ -1,26 +1,27 @@
...
+FROM golang:1.17 as builder
+
+WORKDIR /workspace
+# Copy the Go Modules manifests
+COPY go.mod go.mod
+COPY go.sum go.sum
+# cache deps before building and copying source so that we don't need to re-download as much
+# and so that source changes don't invalidate our downloaded layer
 RUN go mod download
-COPY . .
```
mrbarge, line 7: The removal of this means it won't be pulling in directories such as `pkg`, `config`, `version` and `util` which is what will be causing build problems in the CI space.

Currently it looks like it's only copying `api` and `controllers`.

Should maybe just revert to a copy of `.`.

ravitri: Thanks Matt! Have updated the Dockerfile with the same content as pre-migrated version. That should work mostly.

main.go
```diff
@@ -0,0 +1,341 @@
...
+		setupLog.Error(err, "unable to set up ready check")
+		os.Exit(1)
+	}
+
+	// Define stopCh which we'll use to notify the upgradeConfigManager (and any other routine)
+	// to stop work. This channel can also be used to signal routines to complete any cleanup
+	// work
+	stopCh := signals.SetupSignalHandler()
+
+	upgradeConfigManagerClient, err := client.New(cfg, client.Options{})
```
mrbarge, line 275: As this is setting up its own client it does not have the upgradev1alpha1 types added its scheme. Currently failing to test as a result of that.

I thought it could be supplied with the `mgr.GetClient()` client instead rather than using its own, so I tried that and tested it, but it didn't work. Have not looked deeper into what's happening yet.

mrbarge: Okay, it actually does work to use `mgr.GetClient()` I think, it's just that there's a similar issue elsewhere that was triggering the same problem and I got caught up on assuming that it was still this bit of code. I will comment on the other place it needs addressing.

main.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+
+	// Add the Metrics Service and ServiceMonitor
+	if err := addMetrics(ctx, cfg); err != nil {
+		log.Error(err, "Metrics service is not added.")
+		os.Exit(1)
+	}
+
+	// Add the Custom Metrics Service
+	metricsClient, err := client.New(cfg, client.Options{})
```
mrbarge, line 230: This is the other place where it causes issues due to a lack of type scheme registration.
I'm currently testing by passing a `mgr.GetClient()` into the `NewUpgradeCollector` instead and I think it is working okay (so far)

mrbarge:
Similar to our discussion concerning `rbac-permissions-operator`, the current MUO performs leader-for-life election and the PR loses that behaviour, so something like https://github.com/openshift/rbac-permissions-operator/pull/116 may be needed.

main.go
mrbarge: I ended up going with the separate client and doing the following instead - seems to work ok:
```suggestion
	upgradeConfigManagerClient, err := client.New(cfg, client.Options{Scheme: mgr.GetScheme()})
```

ravitri: Done :+1:

main.go
ravitri: Done :+1:

main.go
```diff
@@ -0,0 +1,346 @@
...
+	log.Info(fmt.Sprintf("Go Version: %s", runtime.Version()))
+	log.Info(fmt.Sprintf("Go OS/Arch: %s/%s", runtime.GOOS, runtime.GOARCH))
+	log.Info(fmt.Sprintf("Version of operator-sdk: v%v", version.SDKVersion))
+}
+
+func main() {
+	var metricsAddr string
+	var enableLeaderElection bool
+	var probeAddr string
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
```
mrbarge, line 121: A minor thing, but this variable is never used anywhere, the `metricsPort` variable is the one that is actually used for setting up the metrics in the code. 

Perhaps it should be changed to `:8383` as a default, and the global `metricsHost` / `metricsPort` are removed, and metricsAddr` is used in the manager setup.

ravitri: Thanks Matt! I have changed this. However, the `metricsPort` is used in `addMetrics()` function to create the service too so I have maintained the `metricsPort` var and used it commonly to initialise the `metricsAddr` var too. Let me know if this seems ok to you.

mrbarge:
I just noticed that the `managed-upgrade-operator-metrics` service isn't being created anymore.

I had a quick attempt at addressing that. Here is one solution, all changes are in `main.go`:

- Add a new kube client parameter to `addMetrics`
```
func addMetrics(ctx context.Context, cl client.Client, cfg *rest.Config) error {
```
It can be supplied the `metricsClient` that is created in `main`:

```
	// Add the Metrics Service and ServiceMonitor
	if err := addMetrics(ctx, metricsClient, cfg); err != nil {
```

- the `addMetrics` function should attempt to create the service after it generates it. But it should also safely handle if the service already existed. (also it should maybe return an error if it couldn't generate it)

```
	service, err := opmetrics.GenerateService(metricsPort, "http-metrics", operatorName+"-metrics", operatorNs, map[string]string{"name": operatorName})
	if err != nil {
		log.Info("Could not generate metrics Service", "error", err.Error())
		return err
	}
	log.Info(fmt.Sprintf("Attempting to create service %s", service.Name))
	err = cl.Create(ctx, service)
	if err != nil {
		if !apierrors.IsAlreadyExists(err) {
			log.Error(err, "Could not create metrics service")
			return err
		} else {
			log.Info("Metrics service already exists, will not create")
		}
	}
```

mrbarge:
Thanks for the fix, service creation is working again now. 
/unhold

mrbarge:
@ravitri for some reason the squash label is not being applied to this PR, maybe it's not possible. Any chance you could squash the commits before we merge? No problem if that turns out to be a hassle though.

main.go
```diff
@@ -0,0 +1,358 @@
...
+		setupLog.Error(err, "unable to start manager")
+		os.Exit(1)
+	}
+
+	// Ensure lock for leader election
+	err = leader.Become(context.TODO(), "managed-upgrade-operator-lock")
+	if err != nil {
+		setupLog.Error(err, "failed to create leader lock")
+		os.Exit(1)
+	}
```
mrbarge, line 176: One small thing here.. how would you feel about changing this so that it only becomes a leader if running on-cluster? Otherwise this block prevents the ability to run locally.

Something like this should maybe work, can you think of anything that might be an issue with it?
```suggestion
	_, err = k8sutil.GetOperatorNamespace()
	if err == nil {
		err = leader.Become(context.TODO(), "managed-upgrade-operator-lock")
		if err != nil {
			setupLog.Error(err, "failed to create leader lock")
			os.Exit(1)
		}
	} else if err != k8sutil.ErrRunLocal {
		setupLog.Error(err, "cannot get operator namespace")
		os.Exit(1)
	}
```

ravitri: Thanks Matt!! I missed to fix this.. I referred to https://github.com/operator-framework/operator-sdk/blob/v0.18.x/pkg/leader/leader.go#L48-L55 and put similar logic in another commit.. Local run skips leader election now.. Let me know if this is ok with you..

mrbarge:
:tada: Thanks for the work here! Let's do this.
/lgtm

PR #335, opened 2022-08-19
adds logic to reinit FIO in fedramp

tonytheleg:
### What type of PR is this?
_(bug/feature/cleanup/documentation/test/refactor)_
feature

The FedRAMP environment utilizes the [File Integrity Operator](https://github.com/openshift/file-integrity-operator) which monitors for changes to files and alerts if monitored files are changed (potential security threat/attack).
Since upgrades can make lots of changes, this PR adds a post-upgrade step to re-initialize FIO in FedRAMP clusters to ensure file changes are updated in its tracking DB.

[OSD-9594](https://issues.redhat.com/browse/OSD-9594)

Currently MUO does not work in FedRAMP (pending other changes to make that possible), so to test that this is working, I did both tests in commercial and changed the logic that the post-upgrade step is checking to see if the cluster is in FedRAMP or not. Since commercial is not impacted by this logic, I figured submit the changes now so if there are changes you'd like to be made, we aren't trying to make them when there is inevitably some crunch time to get this done. Info on the manual tests are available in the Jira story,

- [X] Tested latest changes against a cluster (test results can be find in Jira Card)

### DIFF
--- a/pkg/upgraders/postupgradestep.go
+++ b/pkg/upgraders/postupgradestep.go
@@ -3,13 +3,8 @@ package upgraders
 import (
 	"context"
 	"fmt"
-	"net/url"
-	"strings"
 
 	"github.com/go-logr/logr"
-	"github.com/openshift/managed-upgrade-operator/config"
-	"gopkg.in/yaml.v2"
-	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -23,63 +18,17 @@ const (
 
 var reinitAnnotation = map[string]string{"file-integrity.openshift.io/re-init": ""}
 
-// ugConfigManager stores the configManager section of the upgradeconfig manager configmap
-type ugConfigManager struct {
-	Source     string `yaml:"source"`
-	OcmBaseURL string `yaml:"ocmBaseUrl"`
-}
-
-// ugConfigManagerSpec stores the config.yaml section of the upgradeconfig manager configmap
-type ugConfigManagerSpec struct {
-	Config ugConfigManager `yaml:"configManager"`
-}
-
 // PostUpgradeProcedures are any misc tasks that are needed to be completed after an upgrade has finished to ensure healthy state
 // Currently the only task is to reinit file integrity operator due to changes that come from upgrades
 func (c *clusterUpgrader) PostUpgradeProcedures(ctx context.Context, logger logr.Logger) (bool, error) {
 
-	frCluster, err := c.frClusterCheck(ctx)
-	if err != nil {
-		return false, err
-	}
-	if !frCluster {
+	if !c.config.Environment.IsFedramp() {
 		logger.Info("Non-FedRAMP environment...skipping PostUpgradeFIOReInit ")
 		return true, nil
 	}
-	err = c.postUpgradeFIOReInit(ctx, logger)
+	err := c.postUpgradeFIOReInit(ctx, logger)
 	if err != nil {
 		return false, err
-	}
-	return true, nil
-}
-
-// frClusterCheck checks to see if the upgrading cluster is a FedRAMP cluster to determine if we need to re-init the File Integrity Operator
-func (c *clusterUpgrader) frClusterCheck(ctx context.Context) (bool, error) {
-	ocmConfig := &corev1.ConfigMap{}
-	err := c.client.Get(context.TODO(), client.ObjectKey{Namespace: config.OperatorNamespace, Name: config.ConfigMapName}, ocmConfig)
-
-	if err != nil {
-		return false, fmt.Errorf("failed to fetch %s config map to parse: %v", config.ConfigMapName, err)
-	}
-
-	var cm ugConfigManagerSpec
-	data := fmt.Sprint(ocmConfig.Data["config.yaml"])
-	err = yaml.Unmarshal([]byte(data), &cm)
-
-	if err != nil {
-		return false, fmt.Errorf("failed to parse %s config map for OCM URL: %v", config.ConfigMapName, err)
-	}
-
-	if cm.Config.Source == "OCM" {
-		ocmBaseUrl, err := url.Parse(cm.Config.OcmBaseURL)
-		if err != nil {
-			return false, fmt.Errorf("failed to parse %s config map for OCM URL: %v", config.ConfigMapName, err)
-		}
-		if !strings.Contains(ocmBaseUrl.Host, frOCMBaseDomain) {
-			return false, nil
-		}
-	} else {
-		return false, nil
 	}
 	return true, nil
 }
--- a/pkg/upgraders/postupgradestep.go
+++ b/pkg/upgraders/postupgradestep.go
@@ -11,9 +11,8 @@ import (
 )
 
 const (
-	fioNamespace    string = "openshift-file-integrity"
-	fioObject       string = "osd-fileintegrity"
-	frOCMBaseDomain string = "openshiftusgov.com"
+	fioNamespace string = "openshift-file-integrity"
+	fioObject    string = "osd-fileintegrity"
 )
 
 var reinitAnnotation = map[string]string{"file-integrity.openshift.io/re-init": ""}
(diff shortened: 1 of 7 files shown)


pkg/upgraders/postupgradestep.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
+// frClusterCheck checks to see if the upgrading cluster is a FedRAMP cluster to determine if we need to re-init the File Integrity Operator
+func (c *clusterUpgrader) frClusterCheck(ctx context.Context) (bool, error) {
+	ocmConfig := &corev1.ConfigMap{}
+	err := c.client.Get(context.TODO(), client.ObjectKey{Namespace: config.OperatorNamespace, Name: config.ConfigMapName}, ocmConfig)
```
wking, line 59: not my space, so feel free to ignore me if I'm not making sense.  But wouldn't it make sense to poke the FIO post-update whenever it was installed, on both FedRAMP and otherwise?

tonytheleg: Hey @wking I'm sorry, I'm not sure I follow. Do you mean why not poke FIO no matter what kind of cluster it is, whether FedRAMP or not? This is currently only set for FedRAMP as FIO is not deployed in commercial OSD/ROSA clusters by default where in FedRAMP it is. 

 In FedRAMP its part of the ROSA deployment and used by our team as a requirement of FedRAMP for us to monitor. If a customer decides to use this tool in commercial, while it might be beneficial to add that functionality, they may not all desire us messing with their security configurations? I could see the argument for either way, but this update is targeting specifically only the SREP managed FIO deployments, since customers can deploy to any namespace and name the actual `FileIntegrity` object anything they want. It could be done but not sure if we want that default behavior? Thanks for the input!

wking: >  I could see the argument for either way, but this update is targeting specifically only the SREP managed FIO deployments...

It feels like there should be some way to look at a FileIntegrity resource and decide it its yours or not.  Perhaps by setting [`ownerReferences`][1] for SRE-P-owned resources, and then switching on the presence of those references.  Then "is this my FileIntegrity, or did someone else create a FileIntegrity resource called `osd-fileintegrity` in the `openshift-file-integrity` namespace?" can be answered by inspecting the resource itself.  And really, those naming conventions already seem so SRE-P specific that I would be very surprised to find someone else's FileIntegrity resource by mistake.  But sure, if we think that checking a ConfigMap to decide if we're FedRAMP is more reliable than using something about the FileIntegrity resource itself, then go for it :)

[1]: https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/#owner-references-in-object-specifications

mrbarge: I have to admit this PR gives me a few not-comfortable feelings from the perspective of how this brings in a (indirect, as API types aren't being imported) dependency between the two operators, and maintenance of things like a Fedramp domain name in MUO code.

I'd wish that the FIO had a controller that watched for UpgradeConfigs (perhaps added a finalizer to them, too) and did its thing based on that. If the UpgradeConfig is being deleted and its Status indicates that it completed an upgrade -> reinit whatever needs initializing and remove the finalizer. That divorces MUO's need to have any knowledge.

But given that FIO is an OCP operator and not an OSD/ROSA-specific operator I don't know how likely that change is.

pkg/upgraders/postupgradestep.go
```diff
@@ -0,0 +1,111 @@
...
+		return false, fmt.Errorf("failed to fetch %s config map to parse: %v", config.ConfigMapName, err)
+	}
+
+	var cm ugConfigManagerSpec
+	data := fmt.Sprint(ocmConfig.Data["config.yaml"])
+	err = yaml.Unmarshal([]byte(data), &cm)
+
+	if err != nil {
+		return false, fmt.Errorf("failed to parse %s config map for OCM URL: %v", config.ConfigMapName, err)
+	}
```
mrbarge, line 71: Rather than do yaml wrangling in here, it would be possible to add in something like a `isFedRamp` function into this, which is the part of the code built around extracting upgrade-related config from the ConfigMap:

https://github.com/openshift/managed-upgrade-operator/blob/master/pkg/upgraders/config.go

I am additionally wondering if there are ways we could identify Fedramp clusters that don't involve the combination of the hardcoded domain name and the OCM type. Maybe we could introduce a boolean new field to the `ConfigMap` which identifies FedRamp clusters? And then distribute a Fedramp-specific version of the ConfigMap? Spitballing ideas here.

tonytheleg: I think I like the idea of a FedRAMP boolean in the upgrade configmap. I also think W. Trevor King has a good point here that perhaps using something specific to the FileIntegrity object would be simpler than grabbing the upgrade config, but we'd still be looking at a hard coded value perhaps. We could very easily add a label to the file integrity we create in the [OLM template](https://gitlab.cee.redhat.com/service/file-integrity-operator-catalog/-/blob/master/hack/olm-registry/olm-artifacts-template.yaml#L129) since this is how we're deploying everything into FedRAMP clusters. At that point we can check if the file integrity even exists and if it does check the label and go from there. Not sure which is better?

pkg/upgraders/postupgradestep.go
tonytheleg: I tried pulling in the bits needed from the FIO code base to check for the file integrity API's needed but go versions and dependency differences cause a lot of issues and wouldn't work, so I ended up using the unstructured library. Its been months since I last tried but they are on go 1.16 so not sure it'll work still. I can give it another test if desired.

The issue with adding any MUO specific stuff to FIO is that not just openshift customers may be leveraging FIO since its in OperatorHub. Its a pretty universal tool that could work with any K8s really, so adding something specific to OpenShift might not go well with other users.

wking: are the triggers cluster-scoped or node-scoped?  Because ideally, you'd trigger a node scan after each node change.  [RFE-1925][1] is tracking node-update hooks, which would give you one place to plug in.  On the other hand, it's an RFE because it hasn't been implemented yet.  Tricky problem.

[1]: https://issues.redhat.com/browse/RFE-1925

tonytheleg: When the annotation is applied to the `FileIntegrity` it triggers the daemonset pods to force reinit their database so a single annotation triggers all nodes to update their DB files so that the current file contents are the expected baseline.

pkg/upgraders/postupgradestep.go
tonytheleg: @mrbarge I think I made some of the changes you suggested to the config.go file, if I understood right. I doubt this is working yet as I'm not sure where the config value would be consumed and if it is at the moment, but wanted to make sure i was grokking what you meant. This logic would assume that the muo-config in cluster would look something like
```
data:
  config.yaml: |
   environment:
     fedramp: true
    upgradeType: OSD
    configManager:
      source: OCM
      ocmBaseUrl: ${OCM_BASE_URL}
```
This way when the upgraderConfig is created/parsed we can greatly simplify the check later in post upgrade like mentioned
```
func (c *clusterUpgrader) PostUpgradeProcedures(ctx context.Context, logger logr.Logger) (bool, error) {

	if !c.config.Environment.IsFedramp() {
		logger.Info("Non-FedRAMP environment...skipping PostUpgradeFIOReInit ")
		return true, nil
	}
	err := c.postUpgradeFIOReInit(ctx, logger)
	if err != nil {
		return false, err
	}
	return true, nil
}
```

Does this seem like a more favorable solution that I can work on getting added to submit?

tonytheleg: @mrbarge updates seem to work just fine using this method! This is much nicer. I've added the logs for my test in commercial to the [jira card](https://issues.redhat.com/browse/OSD-9594?focusedCommentId=20850505&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-20850505) using these changes to show that it works. One test with the `managed-upgrade-operator-confg` having fedramp set to false, the other true.

mrbarge: @tonytheleg thanks for making this change, it's exactly what I was thinking of. Looks good and I gave it a quick test successfully as well.
/lgtm

PR #337, opened 2022-09-05
[OSD-12966] Handle error for node drain strategy

a7vicky:
### What type of PR is this?
_(bug)_

Handle error to prevent panic during applying remove finalizer drain strategy.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x141984b]

goroutine 1654 [running]:
github.com/openshift/managed-upgrade-operator/pkg/drain.(*osdDrainStrategy).Execute(0xc000bbe200, 0xc000886c00, {{0x1b78c48, 0xc000a38630}, 0xc000790e90})
        /workdir/pkg/drain/nodeDrainStrategy.go:60 +0x4eb
github.com/openshift/managed-upgrade-operator/controllers/nodekeeper.(*ReconcileNodeKeeper).Reconcile(0xc0000f0930, {0xc000a38540, 0x16e2680}, {{{0x0, 0x18124e0}, {0xc000a1b920, 0x30}}})
```

_Fixes #_[OSD-12966](https://issues.redhat.com/browse/OSD-12966)

### DIFF
--- a/controllers/nodekeeper/nodekeeper_controller.go
+++ b/controllers/nodekeeper/nodekeeper_controller.go
@@ -105,12 +105,13 @@ func (r *ReconcileNodeKeeper) Reconcile(ctx context.Context, request reconcile.R
 		reqLogger.Error(err, "Error while executing drain.")
 		return reconcile.Result{}, err
 	}
+
 	res, err := drainStrategy.Execute(node, reqLogger)
-	for _, r := range res {
-		reqLogger.Info(r.Message)
-	}
 	if err != nil {
 		return reconcile.Result{}, err
+	}
+	for _, r := range res {
+		reqLogger.Info(r.Message)
 	}
 
 	hasFailed, err := drainStrategy.HasFailed(node, reqLogger)
--- a/pkg/pod/pod.go
+++ b/pkg/pod/pod.go
@@ -50,6 +50,7 @@ func DeletePods(c client.Client, logger logr.Logger, pl *corev1.PodList, ignoreA
 			logger.Info(fmt.Sprintf("Applying pod deletion drain strategy to pod %v/%v", p.Namespace, p.Name))
 			err := c.Delete(context.TODO(), &p, options...)
 			if err != nil {
+				logger.Error(err, fmt.Sprintf("failed to delete the pod %v/%v", p.Namespace, p.Name))
 				me = multierror.Append(err, me)
 			} else {
 				podsMarkedForDeletion = append(podsMarkedForDeletion, p.Name)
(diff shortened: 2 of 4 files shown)


controllers/nodekeeper/nodekeeper_controller.go
```diff
@@ -105,13 +105,14 @@ func (r *ReconcileNodeKeeper) Reconcile(ctx context.Context, request reconcile.R
 		reqLogger.Error(err, "Error while executing drain.")
 		return reconcile.Result{}, err
 	}
+
 	res, err := drainStrategy.Execute(node, reqLogger)
-	for _, r := range res {
```
bmeng, line 109: Nice catch! Wondering how would it deal with the error after the for loop...

pkg/pod/pod.go
```diff
@@ -50,6 +50,7 @@ func DeletePods(c client.Client, logger logr.Logger, pl *corev1.PodList, ignoreA
 			logger.Info(fmt.Sprintf("Applying pod deletion drain strategy to pod %v/%v", p.Namespace, p.Name))
 			err := c.Delete(context.TODO(), &p, options...)
 			if err != nil {
+				logger.Error(err, "failed to delete the pod %v/%v", p.Namespace, p.Name)
```
mrbarge, line 53: Either this needs to be wrapped in a sprintf or the key/values need to be specified properly.

```suggestion
				logger.Error(err, fmt.Sprintf("failed to delete the pod %v/%v", p.Namespace, p.Name))
```

or

```suggestion
				logger.Error(err, fmt.Sprintf("failed to delete the pod", "namespace", p.Namespace, "name", p.Name))
```

pkg/pod/pod.go
```diff
@@ -84,6 +85,7 @@ func RemoveFinalizersFromPod(c client.Client, logger logr.Logger, pl *corev1.Pod
 
 			err := c.Update(context.TODO(), &p)
 			if err != nil {
+				logger.Error(err, "failed to remove finalizer from the pod %v/%v", p.Namespace, p.Name)
```
mrbarge, line 88: Same as earlier comment.

```suggestion
				logger.Error(err, fmt.Sprintf("failed to remove finalizer from the pod %v/%v", p.Namespace, p.Name))
```

or

```suggestion
				logger.Error(err, fmt.Sprintf("failed to remove finalizer from the pod", "namespace", p.Namespace, "name", p.Name))
```

pkg/pod/pod.go
a7vicky: Thank you @mrbarge I have applied the suggestion in my last commit.

mrbarge:
Thanks for this fix!
/approve

PR #338, opened 2022-09-06
[OSD-12981] Handle runtime error

a7vicky:
### What type of PR is this?
_(bug)_

_Fixes #_[OSD-12981](https://issues.redhat.com/browse/OSD-12981)

### DIFF
--- a/pkg/collector/collector.go
+++ b/pkg/collector/collector.go
@@ -414,7 +414,9 @@ func collectCondition(c *upgradev1alpha1.UpgradeCondition, promDesc *prometheus.
 
 func (uc *UpgradeCollector) collectSpec(ucfg *upgradev1alpha1.UpgradeConfig, cvV string, ch chan<- prometheus.Metric) error {
 	h := ucfg.Status.History.GetHistory(ucfg.Spec.Desired.Version)
-
+	if h == nil {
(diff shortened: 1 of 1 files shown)


pkg/collector/collector.go
```diff
@@ -414,30 +414,31 @@ func collectCondition(c *upgradev1alpha1.UpgradeCondition, promDesc *prometheus.
 
 func (uc *UpgradeCollector) collectSpec(ucfg *upgradev1alpha1.UpgradeConfig, cvV string, ch chan<- prometheus.Metric) error {
 	h := ucfg.Status.History.GetHistory(ucfg.Spec.Desired.Version)
+	if h != nil {
```
mrbarge, line 417: This is a minor nit, but from a code complexity/readability perspective it looks better if the code exits from the 'unusual' case at the top of the function and then the rest of the code can remain at the base level rather than buried in a nested block.

eg.
```suggestion
	if h == nil {
	   // return the error
	}
	
	// rest of the 'normal' code flow goes here
```

a7vicky: Thank you @mrbarge I have applied the suggested chnage in my last commit.

mrbarge:
Thanks for this fix!
/approve

PR #348, opened 2023-01-19
reset the notification metric on failure, to make sure the MUO could sync with OCM on next schedule

bmeng:
### Description of the issue

Pre:
OCM will cancel the upgrade which is in state pending or scheduled and schedueld at 2h2m ago

Issue:        
1. Failed upgrade will be synced to OCM, and the upgradepolicy will be delete in OCM after 2h2m.

2. New scheduled upgrade will be fetched from OCM on the next MUO sync.

3. The new upgrade will be executed with skip the  `SendStartedNotification` step and the status cannot be synced by to OCM

4. The upgrade could be finished on cluster if the failed condition in step 1 cleared 

5. OCM will delete the upgrade policy after 2h2m if the upgrade on cluster cannot be finished within 2 hours 2 minutes

6. MUO will try to sync to OCM on upgrade completed, but the OCM side upgradepolicy has been deleted

7. The upgradeconfig will be leftover on cluster and the upgrade process cannot be finished.

The reason in 3, that the second scheduled upgrade will skip the `SendStartedNotifcation ` since the `eventmanager.Notify()` function will check the metric value 
of `upgradeoperator_upgrade_notification` and return directly if the value is set. In https://github.com/openshift/managed-upgrade-operator/blob/f38e833a50512e95c6238e79dac683f12302d7b2/pkg/eventmanager/eventmanager.go#L99

The `upgradeoperator_upgrade_notification` was left from the previous upgrade which was not cleaned up.

Also clean up some unused code.

### What type of PR is this?
_bug_

See above

_Fixes #OSD-13795_

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -394,6 +394,7 @@ func (c *Counter) ResetFailureMetrics() {
 		metricUpgradeControlPlaneTimeout,
 		metricUpgradeWorkerTimeout,
 		metricNodeDrainFailed,
+		metricUpgradeNotification,
 	}
 	for _, m := range failureMetricsList {
 		m.Reset()
--- a/pkg/upgraders/notifierstep.go
+++ b/pkg/upgraders/notifierstep.go
@@ -2,9 +2,6 @@ package upgraders
 
 import (
 	"context"
-	"fmt"
-	"time"
-
 	"github.com/go-logr/logr"
 
 	"github.com/openshift/managed-upgrade-operator/pkg/notifier"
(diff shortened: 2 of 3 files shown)


pkg/metrics/metrics.go
```diff
@@ -394,6 +394,7 @@ func (c *Counter) ResetFailureMetrics() {
 		metricUpgradeControlPlaneTimeout,
 		metricUpgradeWorkerTimeout,
 		metricNodeDrainFailed,
+		metricUpgradeNotification,
```
mrbarge, line 397: I think this will work, but we should ensure that this doesn't result in any unintended consequences such as MUO repeatedly trying to notify OCM that the upgrade has failed every reconcile until the UpgradeConfig eventually gets cleaned up.

My quick refresher of the code logic suggests to me that this shouldn't happen, but I'd still like to make sure that we test it and make sure.

pkg/upgraders/notifierstep.go
```diff
@@ -37,34 +34,3 @@ func (c *clusterUpgrader) SendCompletedNotification(ctx context.Context, logger
 	}
 	return true, nil
 }
-
-// UpgradeDelayedCheck checks and sends a notification on a delay to upgrade commencement
-func (c *clusterUpgrader) UpgradeDelayedCheck(ctx context.Context, logger logr.Logger) (bool, error) {
```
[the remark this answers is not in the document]
bmeng: I believe this can be covered in another file:
https://github.com/openshift/managed-upgrade-operator/blob/master/pkg/upgraders/delaystep.go#L15

pkg/metrics/metrics.go
bmeng: It should not I think, based on the controller logic, it should return directly when the upgradeconfig status is `Failed`
https://github.com/openshift/managed-upgrade-operator/blob/f38e833a50512e95c6238e79dac683f12302d7b2/controllers/upgradeconfig/upgradeconfig_controller.go#L227

And also during my testing, I cannot see it is trying to send the notification repeatedly.

pkg/upgraders/notifierstep.go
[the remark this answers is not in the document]
mrbarge: Oh my bad, I didn't realize it was being defined in two spots. Thanks for the clarification.

pkg/metrics/metrics.go
mrbarge: Thanks for confirming that it's been checked/tested, that's all I really wanted to know here.

mrbarge:
Thanks for the fix here!
/lgtm

PR #349, opened 2023-01-25
MUO should accept conditionalUpgrades

jewzaam:
### What type of PR is this?
bug

Permit MUO to upgrade to any conditional edge.  This risk is managed externally and MUO should accept any version set in an upgrade policy.

https://issues.redhat.com/browse/OSD-14971


fahlmant (approved):
/approve
/lgtm

PR #360, opened 2023-04-05
Update Makefile

a7vicky:
This PR fix `make run ` to start operator locally for development purpose


mrbarge:
@a7vicky this'll require either rebasing off master or running `boilerplate/_lib/container-make validate` and committing the file in the repo that is changed as a result (probably `Dockerfile.olm-registry`).

Looks like that file changed as part of a recent dependabot update. OAO had the same issue with Wenqi's PR for some odd reason.

PR #390, opened 2023-10-25
Increase ErrClusterIDNotFound verbosity

tnierman:
### What type of PR is this?
_cleanup_

Increases the verbosity of the `ErrClusterIDNotFound` error message. 

Anecdotally, this error is thrown most often when MUO is unauthorized to lookup clusters in OCM. As it stands, the error gives no indication that the problem may be a pull-secret/authentication one; suggesting this within the operator short-circuits what can be otherwise lengthy investigations.

### DIFF
--- a/pkg/ocm/client.go
+++ b/pkg/ocm/client.go
@@ -32,7 +32,7 @@ var log = logf.Log.WithName("ocm-client")
 
 var (
 	// ErrClusterIdNotFound is an error describing the cluster ID can not be found
-	ErrClusterIdNotFound = fmt.Errorf("cluster ID can't be found")
+	ErrClusterIdNotFound = fmt.Errorf("OCM did not return a valid cluster ID: pull-secret may be invalid")
 )
 
 // OcmClient enables an implementation of an ocm client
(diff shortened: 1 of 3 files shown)


mjlshen (approved):
/lgtm

pkg/ocm/client.go
```diff
@@ -32,7 +32,7 @@ var log = logf.Log.WithName("ocm-client")
 
 var (
 	// ErrClusterIdNotFound is an error describing the cluster ID can not be found
-	ErrClusterIdNotFound = fmt.Errorf("cluster ID can't be found")
+	ErrClusterIdNotFound = fmt.Errorf("OCM did not return a valid cluster ID: pull-secret may be invalid")
```
a7vicky, line 35: Thank you for The PR. Thinking if we could add Cluster's owner-banned scenario in the log info for more clarity. ref: https://access.redhat.com/articles/1340183
```suggestion
	ErrClusterIdNotFound = fmt.Errorf("OCM did not return a valid cluster ID: pull-secret may be invalid OR cluster's owner is disabled/banned in OCM")
```

PR #394, opened 2023-11-16
Add in-cluster OCM agent proxy for OCM requests

hectorakemp:
### What type of PR is this?
_(feature)_

Adds OCM agent handler for the MUO OCM client. 
If we set the OCM baseURL like so: 
```yaml
configManager:
  source: OCM
  ocmBaseUrl: http://ocm-agent.openshift-ocm-agent-operator.svc.cluster.local:8081
  watchInterval: 60
```
It will activate the `pkg/ocmagent/client.go` handler, which uses the new OCM agent upgrade policies/clusters endpoints (https://github.com/openshift/ocm-agent/pull/69 and https://github.com/openshift/ocm-agent/pull/64) and return identical responses to the 'regular' `pkg/ocm/client`, as it implements the `ocmAgent` interface.

_Fixes #_
https://issues.redhat.com/browse/OSD-19698


hectorakemp:
Needs boilerplate update (https://github.com/openshift/managed-upgrade-operator/pull/397/files) to pass CI

pkg/ocmagent/client.go
```diff
@@ -0,0 +1,170 @@
...
+func (ort *ocmRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	transport := http.Transport{
+		TLSHandshakeTimeout: time.Second * 5,
+	}
+	return transport.RoundTrip(req)
+}
+
+// Read cluster info from OCM
+func (s *ocmClient) GetCluster() (*ocm.ClusterInfo, error) {
+	csUrl, err := url.Parse(s.ocmBaseUrl.String())
```
bmeng, line 54: We have this variable in each function, can we use the same name for all?

bmeng: Or can we move the validating outside these functions?

pkg/ocmagent/client.go
```diff
@@ -0,0 +1,170 @@
...
+		Kind: "UpgradePolicyList",
+		Page: 1,
+		Size: int64(len(*upgradeResponse)),
+		Total: int64(len(*upgradeResponse)),
+		Items: *upgradeResponse,
+	}, nil
+}
+
+// Send a notification of state
+func (s *ocmClient) SetState(value string, description string, policyId string, clusterId string) error {
```
bmeng, line 115:
```suggestion
func (s *ocmClient) SetClusterUpgradePolicyState(value string, description string, policyId string, clusterId string) error {
```

pkg/ocmagent/client.go
```diff
@@ -0,0 +1,170 @@
...
+var log = logf.Log.WithName("ocm-client")
+
+var (
+	// ErrClusterIdNotFound is an error describing the cluster ID can not be found
+	ErrClusterIdNotFound = fmt.Errorf("OCM did not return a valid cluster ID: pull-secret may be invalid OR cluster's owner is disabled/banned in OCM")
+)
+
+type ocmClient struct {
+	// Cluster k8s client
+	client client.Client
```
bmeng, line 36: Do we need the kube client here?

pkg/ocmagent/client.go
```diff
@@ -0,0 +1,170 @@
...
+	UPGRADEPOLICIES_V1_PATH = "upgrade_policies"
+	// STATE_V1_PATH sub-path to the policy state service
+	STATE_V1_PATH = "state"
+)
+
+var log = logf.Log.WithName("ocm-client")
+
+var (
+	// ErrClusterIdNotFound is an error describing the cluster ID can not be found
+	ErrClusterIdNotFound = fmt.Errorf("OCM did not return a valid cluster ID: pull-secret may be invalid OR cluster's owner is disabled/banned in OCM")
```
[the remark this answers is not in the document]
hectorakemp: Same as above, copied from the original ocm/client package. Will remove it.

hectorakemp (commented):
Thanks for the review @bmeng ! Will address the comments :)

hectorakemp:
Tested successfully with a fresh cluster that's upgraded from 4.14.1->3:
https://gitlab.cee.redhat.com/-/snippets/7451
https://redhat-internal.slack.com/archives/C0326L38PEH/p1701314925654959

pkg/ocmprovider/ocmprovider.go
```diff
@@ -12,6 +12,7 @@ import (
 
 	upgradev1alpha1 "github.com/openshift/managed-upgrade-operator/api/v1alpha1"
 	"github.com/openshift/managed-upgrade-operator/pkg/ocm"
+	"github.com/openshift/managed-upgrade-operator/pkg/ocmagent"
```
bmeng, line 15: Can we build this as a separated provider, called `ocmagentprovider`? 
So that we don't need to change anything within the existing `ocmprovider` and we can switch between them easily.

bmeng: And them, we can add another condition here 
https://github.com/openshift/managed-upgrade-operator/blob/master/pkg/specprovider/specprovider.go#L44

bmeng: but TBH, I am not sure if it is better or not, if the code are almost the same, there might be just a copy/paste of code blocks.

weherdh:
@hectorakemp I just have one question, since currently from the [mucm](https://github.com/openshift/managed-cluster-config/blob/master/deploy/managed-upgrade-operator-config/10-managed-upgrade-operator-configmap.yaml#L11), the `ocmBaseUrl` value would be passed from the `app-interface`'s deployment([Example](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/products/osdv4/environments/production-hivep01ue1.yml#L19)), so we gonna assign the value in MCC with `http://ocm-agent.openshift-ocm-agent-operator.svc.cluster.local:8081` to make this take effect? Otherwise, nothing impacts, right?

pkg/ocmprovider/ocmprovider.go
hectorakemp: I don't mind either way - whichever one you think is better, I am happy to do.

hectorakemp:
> @hectorakemp I just have one question, since currently from the [mucm](https://github.com/openshift/managed-cluster-config/blob/master/deploy/managed-upgrade-operator-config/10-managed-upgrade-operator-configmap.yaml#L11), the `ocmBaseUrl` value would be passed from the `app-interface`'s deployment([Example](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/products/osdv4/environments/production-hivep01ue1.yml#L19)), so we gonna assign the value in MCC with `http://ocm-agent.openshift-ocm-agent-operator.svc.cluster.local:8081` to make this take effect? Otherwise, nothing impacts, right?

Correct @weherdh @bmeng - either we do it that way,  or we use a different provider (as @bmeng mentioned in his review comments). I don't mind either way, as they both end up with the same result.

hectorakemp:
@bmeng here is the MCC PR: 
https://github.com/openshift/managed-cluster-config/pull/1948

hectorakemp:
And OCM agent: https://github.com/openshift/ocm-agent/pull/71
They require this PR to be merged first

PR #416, opened 2024-04-02
Handle default worker machineset not found error

a7vicky:
### What type of PR is this?
_(bug)_

It gracefully handles the error response Notfound for the reource machineset with label `hive.openshift.io/machine-pool": "worker"` and bypasses the upgrade without capacity reservation.

_Fixes #_ [OSD-21898](https://issues.redhat.com/browse/OSD-21898)


a7vicky:
/hold
Pending test

pkg/scaler/machineSetScaler.go
```diff
@@ -38,9 +39,14 @@ func (s *machineSetScaler) CanScale(c client.Client, logger logr.Logger) (bool,
 		client.MatchingLabels{"hive.openshift.io/machine-pool": "worker"},
 	}...)
 	if err != nil {
-		logger.Error(err, "failed to get original machinesets")
-		return false, err
+		if apierrors.IsNotFound(err) {
+			logger.Error(err, "failed to get original machinesets")
```
mjlshen, line 43:
```suggestion
			logger.Info(err, "failed to get original machinesets")
```
If it's not an error anymore, I would recommend downgrading this to an info or warning level. I think info makes sense because not having the default machineset isn't a bug/misconfiguration.

ravitri:
@a7vicky - Thanks a lot ! 

Can we add test and doc change for this PR too? I think it will be quite helpful.

a7vicky:
> @a7vicky - Thanks a lot !
> 
> Can we add test and doc change for this PR too? I think it will be quite helpful.

Thanks, @ravitri for the review. I could see the test cases are already converted [here](https://github.com/openshift/managed-upgrade-operator/blob/master/pkg/scaler/scaling_test.go).
For now, I have updated the FAQ doc for the same.

a7vicky:
This quick fix is not needed. It turns out to be the issue with OSD/ROSA cluster which was not running the latest MUO version which contains the logic to skip the capacity reservation stage if the default machineset doesn't exist in the cluster.
Closing this MR for now and we will take this ahead in MUO maintenance epic.

PR #417, opened 2024-04-04
Setting metric UpdateMetricNodeDrainFailed only for the upgrading node

a7vicky:
### What type of PR is this?
_(refactor)_

Setting metric `UpdateMetricNodeDrainFailed` only for the node going through upgrade

_Fixes #_ [OSD-20246](https://issues.redhat.com/browse/OSD-20246)


pkg/machinery/node.go
```diff
@@ -29,3 +29,12 @@ func (m *machinery) IsNodeCordoned(node *corev1.Node) *IsCordonedResult {
 		AddedAt:    cordonAddedTime,
 	}
 }
+
+// IsNodeUpgrading returns bool
+func (m *machinery) IsNodeUpgrading(node *corev1.Node) bool {
+	if node.Annotations["machineconfiguration.openshift.io/currentConfig"] != node.Annotations["machineconfiguration.openshift.io/desiredConfig"] {
```
ravitri, line 35: I think it's better to check if the `machineconfiguration.openshift.io/state` annotation value is `Working` which AFAIK implies that the configs are validated and there are no issues found at MCD level which otherwise would cause the state to be `Degraded`. 

What I mean is, when the current and desired config is not the same, the state annotation could either be `Working` or `Degraded`. But if for any reason MCO/MCD is not able to reboot the node at all since node is Degraded, at this point SRE intervention is required before it is qualified for upgrade. 

If it helps, we can run a quick test with two scenarios:

1.  `oc debug` into a node and intentionally change a file which is managed by MCO and let the node become `Degraded`. At this time, upgrade should get stuck or SRE should get some alert when the current and desired config is not the same but the node update is stuck too.
2. Second test is the usual one where node is all fine and the usual upgrade at MCO level is marked by state being `Working`. 

(When I say state, I mean `machineconfiguration.openshift.io/state` annotation)

Let me know your thoughts on same.

a7vicky: Thanks, Ravi for the review. I updated the requested change in my latest commit

controllers/nodekeeper/nodekeeper_controller.go
```diff
@@ -121,7 +121,10 @@ func (r *ReconcileNodeKeeper) Reconcile(ctx context.Context, request reconcile.R
 		}
 		if hasFailed {
 			reqLogger.Info(fmt.Sprintf("Node drain timed out %s. Alerting.", node.Name))
-			metricsClient.UpdateMetricNodeDrainFailed(node.Name)
+			// Set metric only for the node going through upgrade
+			if r.Machinery.IsNodeUpgrading(node) {
```
bmeng, line 125: I am not quite sure if we add this condition check could fix the issue, as the existing logic, if we went to nodeDrain stage, the node should be under upgrade anyway.
Last time I saw the issue, it is not where we set the metric, but the metric was not unset after the node upgrade completed, even the node is gone, the alerts for that node is still firing.

a7vicky: @bmeng There are multiple edge cases associated with this metric. In this case, we want to overcome the scenario where the drain strategy runs for the node that is marked unschedulable because the MHC detected node is unhealthy and deleted the machine object. Because the node is unschedulable as per MUO logic it detects the node is going through an upgrade, then runs a drain strategy for that node and sets this metric. MHC deletes the node and the metric never resets for the node. Therefore we added a logic to only set metrics for the node that is upgrading using **IsNodeUpgrading**

PR #421, opened 2024-04-29
Refactor healthchecks for clusteroperator and critical alerts

ravitri:
### What type of PR is this?
_(refactor)_

This PR splits the health checks regarding critical alerts and clusteroperators to their own files such that pre and post healthchecks can simply include/exclude any required healthchecks and each healthcheck can have it's individual scope and testing. 

_Fixes #_ [OSD-22082](https://issues.redhat.com/browse/OSD-22082)

- There were certainly multiple approaches in refactoring the healthchecks and it can be improved further too.
- Moving healthchecks to it's own package is a decision we might take in future based on requirements but isn't immediately required.
- Existing healthchecks for critical alerts and clusteroperators will be shortly improved further with servicelogs and their own metrics.
- Currently, metric is only set for critical alerts check but not clusteroperator healthcheck failure based on how the state was before refactoring. We will be updating this ahead as separate PR.
- 


pkg/upgraders/healthcheckstep.go
```diff
@@ -23,80 +20,29 @@ func (c *clusterUpgrader) PreUpgradeHealthCheck(ctx context.Context, logger logr
...
 	}
 
-	c.metrics.UpdateMetricClusterCheckSucceeded(c.upgradeConfig.Name)
-	return true, nil
-}
-
-// PostUpgradeHealthCheck performs cluster healthy check
-func (c *clusterUpgrader) PostUpgradeHealthCheck(ctx context.Context, logger logr.Logger) (bool, error) {
-	ok, err := performClusterHealthCheck(c.metrics, c.cvClient, c.config, logger)
+	ok, err = ClusterOperators(c.metrics, c.cvClient, c.upgradeConfig, logger)
```
tkong-redhat, line 28: No problem with the code. But on the logic point of view, do we need to check all the healthy condition?
In this implementation, if there is any critical alert, code will return then there is no check on the cluster operators.

ravitri: Thanks Tony! Yeah, nice point. For now, the checks are blocking in nature and we are maintaining that. So if there's any health check failing then upgrade will not proceed. However, we will classify these failures better and categorize into blocking/non-blocking alerts/issues and notify accordingly. For now, the health check continues to remain similar as in https://github.com/openshift/managed-upgrade-operator/blob/master/pkg/upgraders/healthcheckstep.go#L51-L101 in this refactoring PR. In future PR we will gather these failures together and notify once. Let me know if this answers your query.

tkong-redhat: I understand that any block alerts/issues will make the upgrade not be processed. My question is there is a chance that both blocking alert and blocking degraded cluster operator exist. In this case, the alert check will fail then code returns, the degraded operators will not be checked. Then only alert failure has been notified. Customer may not notice there is also blocking degraded cluster operator.
Let me know if I misunderstand something.

tkong-redhat (approved):

PR #426, opened 2024-05-29
Integrating Pre-HealthCheck with UpgradeConfig controller

ravitri:
### What type of PR is this?
_(feature)_

This feature PR enables running the Pre-HealthCheck when an upgrade is scheduled in advance (atleast more than 2 hours by default) and notifies the user/admin of any healthcheck failures before the upgrade starts. It would help the users/admin to take action on failing healthchecks (as relevant) which will ensure higher success rate of upgrades. The requirement is to the Pre-HealthCheck once and only once before the upgrade starts but once the upgrade is scheduled. Thus, it's implemented in the `New` Upgrade phase.

At a high level following are the noteworthy points regarding the change:
- Previously, the `New` and `Pending` state were treated the same but now, they are split into separate states. This is a crucial change which requires additional tests/validation.
- When an upgrade is scheduled, the upgrade phase is set to `New` first.
- In the `New` state, a Pre-HealthCheck is run **only** if the time to upgrade is more than 2 hours. This value of 2 hours is defined by new field in `UpgradeConfig` API called `HealthCheckDuration`. Currently, this won't be exposed to OCM API and will always be 2 hours unless an upgrade is run in LOCAL mode.
- Regardless of whether the Pre-HealthCheck will pass or fail, the upgrade state is moved to "Pending" state in one minute reconcile time where it will wait for the upgrade schedule time to proceed with upgrade.

_Fixes #_ [OSD-22454](https://issues.redhat.com/browse/OSD-22454)

Following tests were done so far:
- Scheduling upgrade from OCM in advance and checking if Pre-HealthCheck is run = PASS
- Scheduling upgrade from OCM immediately and checkig if Pre-HealthCheck is skipped = PASS
- Scheduling an upgrade and cancelling it = PASS
- Local run of upgrade passes as well with tests = PASS
- Checking metrics for health check run = PASS

Some additional noteworthy points:
- Scheduling an upgrade and cancelling it works. However, this needs more testing and validation. The UpgradeConfig remains in Pending state before the next OCM poll is going to sync and get the UpgradeConfig removed. The end state is fine but the interim state till OCM sync happens doesn't seem right. Will need to fix this separately.
- The default value for `HealthCheckDuration` needs some more testing as well. For now, the value is enforced via a separate function. Will need to test this better.

### DIFF
--- a/api/v1alpha1/upgradeconfig_types.go
+++ b/api/v1alpha1/upgradeconfig_types.go
@@ -35,6 +35,14 @@ type UpgradeConfigSpec struct {
 
 	// Specify if scaling up an extra node for capacity reservation before upgrade starts is needed
 	CapacityReservation bool `json:"capacityReservation,omitempty"`
+
+	// +kubebuilder:validation:Minimum:=2
+	// +kubebuilder:validation:Default:=2
+	// +kubebuilder:default:=2
+	// +kubebuilder:validation:Optional
+	// HealthCheckDuration specifies minimum duration of 2 hours before upgrade time to run healthcheck.
+	// If an upgrade is scheduled within 2 hours from now, healthcheck won't run.
+	HealthCheckDuration int32 `json:"HealthCheckDuration,omitempty"`
 }
 
 // UpgradeConfigStatus defines the observed state of UpgradeConfig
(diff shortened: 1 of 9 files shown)


ravitri:
/hold

Adding hold to update boilerplate to allow validate test to pass. The boilerplate update will need to be a separate PR. Will rebase this PR then.

ravitri:
PR https://github.com/openshift/managed-upgrade-operator/pull/425 is merged now so validate should ideally pass

ravitri:
/unhold

Adding unhold since the PR tests are passing now

shitaljante:
I would like to verify state transition cases for "New" and "Pending" through e2e tests, I will add those in QE doc.  
Otherwise changes in PR looks good to me. 
/lgtm

PR #428, opened 2024-05-29
No worker machinepool HealthCheck

a7vicky:
### What type of PR is this?
_(feature)_

- Send notification during upgrade in case MUO is unable to perform compute capacity reservation because the default worker machine pool is deleted 
- As part of the health check it verify the default worker machine pool exists in the cluster or not and returns the value accordingly
- Set metrics if the health check fails

_Fixes #_[OSD-22459](https://issues.redhat.com/browse/OSD-22459)

### DIFF
--- a/pkg/upgraders/healthcheckstep.go
+++ b/pkg/upgraders/healthcheckstep.go
@@ -7,6 +7,7 @@ import (
 	"github.com/go-logr/logr"
 
 	upgradev1alpha1 "github.com/openshift/managed-upgrade-operator/api/v1alpha1"
+	"github.com/openshift/managed-upgrade-operator/pkg/metrics"
 )
 
 // PreUpgradeHealthCheck performs cluster healthy check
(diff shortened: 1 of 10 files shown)


shitaljante:
lgtm

pkg/upgraders/healthcheckstep.go
```diff
@@ -40,6 +41,18 @@ func (c *clusterUpgrader) PreUpgradeHealthCheck(ctx context.Context, logger logr
 		}
 		return false, err
 	}
+
+	if c.upgradeConfig.Spec.CapacityReservation {
+		ok, err := c.scaler.CanScale(c.client, logger)
+		if !ok {
+			c.metrics.UpdateMetricHealthcheckFailed(c.upgradeConfig.Name, metrics.DefaultWorkerMachinepoolNotFound)
```
ravitri, line 48: So currently, there's no notification (`c.notifier.Notify(notifier.MuoStateHealthCheck)`) sent for this failure in pre-healthcheck. Is that missed?

ravitri:
/lgtm

Thanks a lot Abhishek!

PR #432, opened 2024-06-06
Osd 22457 add manual cordon node check

tkong-redhat:
Add pre-flight health check for manually cordoned worker node

### DIFF
--- a/pkg/upgraders/healthcheck_nodes.go
+++ b/pkg/upgraders/healthcheck_nodes.go
@@ -26,7 +26,7 @@ func ManuallyCordonedNodes(metricsClient metrics.Metrics, machinery machinery.Ma
 	err := c.List(context.TODO(), nodes, cops)
 	if err != nil {
 		logger.Info("Unable to fetch node list")
-		metricsClient.UpdateMetricHealthcheckFailed(ug.Name, metrics.ClusternNodesManuallyCordoned)
+		metricsClient.UpdateMetricHealthcheckFailed(ug.Name, metrics.ClusterNodeQueryFailed)
 		return false, err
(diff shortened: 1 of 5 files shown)


pkg/upgraders/healthcheck_nodes.go
```diff
@@ -0,0 +1,53 @@
...
+	cops := &client.ListOptions{
+		Raw: &metav1.ListOptions{
+			LabelSelector: "node-role.kubernetes.io/worker, !node-role.kubernetes.io/infra",
+		},
+	}
+	// Get the list of worker nodes
+	err := c.List(context.TODO(), nodes, cops)
+	if err != nil {
+		logger.Info("Unable to fetch node list")
+		metricsClient.UpdateMetricHealthcheckFailed(ug.Name, metrics.ClusternNodesManuallyCordoned)
```
ravitri, line 29: This metric is a bit misleading. If for some reason the API itself is not available, we need to set a different reason for failure like "healthcheck_node_query_failed" like in https://github.com/openshift/managed-upgrade-operator/blob/master/pkg/upgraders/healthcheck_alerts.go#L33 for metric query failing. Here it's possible that all the kubelets in the nodes are fine but due to API issue the list of nodes doesn't work.

Mainly, the "reason" label for healthcheck metric should correctly reflect where the failure happened.

tkong-redhat: Thank you, Ravi. Make sense, I'll make the update.

ravitri:
/label tide/merge-method-squash
/lgtm
/approve

Thanks a lot Tony! :+1:

PR #438, opened 2024-06-20
Reset the NodeDrainFailed metric when the node deletion timestamp is set

chamalabey:
### What type of PR is this?
_bug_

_Need to add a validation check to the Nodekeeper controller to verify, if the node has a deletion timestamp set then reset the metric and never set it again. This will make sure that we will not get an alert UpgradeNodeDrainFailedSRE once the node is deleted by MHC_
_Fixes #OSD-20246_

- [x] All the unit tests are passing without errors


ravitri:
Thanks for the PR Chamal ! The PR seems fine but can you add the unit test for it?

a7vicky:
Thanks for the PR @chamalabey 
/lgtm
/approve
Feel free to remove the hold.

PR #439, opened 2024-06-21
Enabling ServiceLog messages for Upgrade phases

ravitri:
### What type of PR is this?
_(feature)_

As part of MUO we want users to have more visibility on the progress of upgrades when a upgrade phase is starting/finished. This PR introduces such notifications via ServiceLogs when:

1. Upgrade starts (i.e control and worker plane upgrade start together)
2. Control Plane upgrade finishes
3. Worker Plane upgrade finishes

Currently, respective MUO states have been added and mapped to respective ServiceLogs similarly to the MUO state mapping to OCM states mapping.

_Fixes #_ [OSD-18464](https://issues.redhat.com/browse/OSD-18464)

- Separate PR coming up soon with more unit tests and documentation updates. For now, this PR would unblock other features to be added regarding notifications.
- Full upgrade is tested and the notifications in OCM are there as expected. Please refer to [OSD-18464](https://issues.redhat.com/browse/OSD-18464) for details.


pkg/eventmanager/eventmanager.go
```diff
@@ -127,6 +140,12 @@ func (s *eventManager) Notify(state notifier.MuoState) error {
 		description = createFailureDescription(uc)
 	case notifier.MuoStateHealthCheck:
 		description = fmt.Sprintf(UPGRADE_HEALTHCHECK_DELAY_DESC, uc.Spec.Desired.Version)
+	case notifier.MuoStateControlPlaneUpgradeStartedSL:
+		description = fmt.Sprintf(UPGRADE_CONTROL_PLANE_STARTED_DESC, uc.Spec.Desired.Version)
+	case notifier.MuoStateControlPlaneUpgradeFinishedSL:
+		description = fmt.Sprintf(UPGRADE_CONTROL_PLANE_FINISHED_DESC, uc.Spec.Desired.Version)
+	case notifier.MuoStateWorkerPlaneUpgradeFinishedSL:
```
tkong-redhat, line 147: I'm OK with the current implementation here. I just have one question that since we have control plane upgrade started notification, will customer complain that they didn't see worker plane upgrade start notification?

pkg/notifier/notifier.go
```diff
@@ -27,16 +27,19 @@ type NotifierBuilder interface {
...
+	MuoStateStarted                       MuoState = "StateStarted"
+	MuoStateCompleted                     MuoState = "StateCompleted"
+	MuoStateDelayed                       MuoState = "StateDelayed"
+	MuoStateFailed                        MuoState = "StateFailed"
+	MuoStateCancelled                     MuoState = "StateCancelled"
+	MuoStateScheduled                     MuoState = "StateScheduled"
+	MuoStateScaleSkipped                  MuoState = "StateScaleSkipped"
+	MuoStateSkipped                       MuoState = "StateSkipped"
+	MuoStateHealthCheck                   MuoState = "StateHealthCheck"
+	MuoStateControlPlaneUpgradeStartedSL  MuoState = "StateControlPlaneStartedSL"
```
tkong-redhat, line 40: There are 3 new states introduced here, will validateStateTransition be updated?

pkg/eventmanager/eventmanager.go
ravitri: Yeah, agreed we need to make it more clear. The SL was already fixed but I'll add another quick fix on the vars side as well.

pkg/notifier/notifier.go
ravitri: No. `validateStateTransition` is for the mapping between MUO states and OCM states. But for the notifications we only use MUO states along with SL states so won't require to add it there.

tkong-redhat: Understood

PR #440, opened 2024-06-26
Osd 24234 use bound API token for prometheus-k8s in metric pkg

tkong-redhat:
Using bound API token for prometheus-k8s in metric pkg due to 4.16 stops automatically generating serviceaccount token.


shitaljante:
Changes looks good. Thanks Tony!

pkg/metrics/metrics.go
```diff
@@ -565,27 +576,31 @@ func (c *Counter) Query(query string) (*AlertResponse, error) {
...
-	token := ""
-	for _, secret := range sl.Items {
-		if strings.HasPrefix(secret.Name, "prometheus-k8s-token") {
-			tokendata := secret.Data[corev1.ServiceAccountTokenKey]
-			token = string(tokendata)
-			break
-		}
+	k8sClient, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return nil, fmt.Errorf("unable to creat kubernets clientSet to: %w", err)
```
ravitri, line 587:
```suggestion
		return nil, fmt.Errorf("unable to create kubernetes clientSet to: %w", err)
```

pkg/metrics/metrics.go
```diff
@@ -565,27 +576,31 @@ func (c *Counter) Query(query string) (*AlertResponse, error) {
...
-		return nil, fmt.Errorf("failed to find token secret for prometheus-k8s SA")
-	}
+	return k8sClient, nil
+}
 
-	return &token, nil
+// RequestPrometheusServiceAccountAPIToken returns a time-bound (1hr) API token for the prometheus service account.
+func RequestPrometheusServiceAccountAPIToken(kc kubernetes.Clientset) (string, error) {
+	expirationSeconds := int64(1 * time.Hour / time.Second)
+	req, err := kc.CoreV1().ServiceAccounts(MonitoringNS).CreateToken(context.TODO(), promApp,
```
ravitri, line 596: As discussed, we should have the legacy mechanism of referring the prometheus token too if it is present. So that will keep the mechanism for clusters < 4.16 intact and 4.16 onwards it will request for new short lived token as required.

pkg/metrics/metrics.go
tkong-redhat: Updated

pkg/metrics/metrics.go
tkong-redhat: Updated

pkg/metrics/metrics.go
```diff
@@ -582,12 +586,49 @@ func prometheusToken(c client.Client) (*string, error) {
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to find token secret for prometheus-k8s SA due to %v", err)
+		}
 	}
 
 	return &token, nil
 }
 
+func NewKubernetesClient() (*kubernetes.Clientset, error) {
+	config, err := rest.InClusterConfig()
```
ravitri, line 600: Can we have a look at https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.17.0/pkg/client#SubResourceClientConstructor.SubResource to leverage controller-runtime client to make the call to create the token?

pkg/metrics/metrics.go
```diff
@@ -582,12 +584,31 @@ func prometheusToken(c client.Client) (*string, error) {
...
+			return nil, fmt.Errorf("failed to find token secret for prometheus-k8s SA due to %v", err)
+		}
 	}
 
 	return &token, nil
 }
 
+// RequestPrometheusServiceAccountAPIToken returns a time-bound (24hr) API token for the prometheus service account.
+func RequestPrometheusServiceAccountAPIToken(c client.Client) (string, error) {
+	expirationSeconds := int64(10 * time.Minute / time.Second)
```
ravitri, line 599: Can we change this to 24 hours again or do you intend it to be 10 minutes ahead? If 10 mins then will need to update the function comment as well.

pkg/metrics/metrics.go
```diff
@@ -582,12 +584,31 @@ func prometheusToken(c client.Client) (*string, error) {
...
+	sa := &corev1.ServiceAccount{ObjectMeta: metav1.ObjectMeta{Namespace: MonitoringNS, Name: promApp}}
+	token := &authenticationv1.TokenRequest{
+		Spec: authenticationv1.TokenRequestSpec{ExpirationSeconds: &expirationSeconds},
+	}
+
+	err := c.SubResource("token").Create(context.TODO(), sa, token)
+	if err != nil {
+		return "", fmt.Errorf("unable to get an API token for the %s service account in the %s namespace: %w", promApp, MonitoringNS, err)
```
ravitri, line 607:
```suggestion
		return "", fmt.Errorf("unable to create an API token for the %s service account in the %s namespace: %w", promApp, MonitoringNS, err)
```

pkg/metrics/metrics.go
```diff
@@ -582,12 +584,31 @@ func prometheusToken(c client.Client) (*string, error) {
 	}
 
 	if len(token) == 0 {
-		return nil, fmt.Errorf("failed to find token secret for prometheus-k8s SA")
+		// If serviceaccount token doesn't exist, try to use token request to get a token
+		token, err = RequestPrometheusServiceAccountAPIToken(c)
+		if err != nil {
+			return nil, fmt.Errorf("failed to find token secret for prometheus-k8s SA due to %v", err)
```
ravitri, line 590:
```suggestion
			return nil, fmt.Errorf("failed to request token for prometheus-k8s SA due to %v", err)
```

pkg/maintenance/alertmanagerMaintenance.go
```diff
@@ -94,13 +94,21 @@ func getAuthentication(c client.Client) (runtime.ClientAuthInfoWriter, error) {
...
 		if strings.Contains(s.Name, alertManagerServiceAccountName+"-token") {
 			token = string(s.Data["token"])
 		}
 	}
 
+	if len(token) == 0 {
+		// If serviceaccount token doesn't exist, try to use token request to get a token
+		token, err = metrics.RequestPrometheusServiceAccountAPIToken(c)
+		if err != nil {
+			return nil, fmt.Errorf("failed to find token secret for prometheus-k8s SA due to %v", err)
```
ravitri, line 108:
```suggestion
			return nil, fmt.Errorf("failed to request token for prometheus-k8s SA due to %v", err)
```

hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -108,6 +108,14 @@ objects:
...
             - list
             - watch
+          - apiGroups:
+            - ""
+            resources:
+            - serviceaccounts/token
+            verbs:
+            - create
+            - update
+            - patch
```
ravitri, line 118: Just wanted to confirm if we need the patch and update verbs as well?

hack/olm-registry/olm-artifacts-template.fedramp.yaml
```diff
@@ -116,6 +116,14 @@ objects:
...
             - list
             - watch
+          - apiGroups:
+            - ""
+            resources:
+            - serviceaccounts/token
+            verbs:
+            - create
+            - update
+            - patch
```
ravitri, line 126: Just wanted to confirm if we need the update and patch verbs as well?

tkong-redhat: No, I'll update this

pkg/metrics/metrics.go
tkong-redhat: Yes, the time will be changed to 24 hours. 10 mins here just for testing.

hack/olm-registry/olm-artifacts-template.fedramp.yaml
tkong-redhat: Removed unnecessary verbs

hack/olm-registry/olm-artifacts-template.yaml
tkong-redhat: Removed unnecessary verbs

pkg/metrics/metrics.go
tkong-redhat: Updated to use controller runtime

pkg/metrics/metrics.go
tkong-redhat: Change back to 24 hours

pkg/metrics/metrics.go
tkong-redhat: Updated

pkg/metrics/metrics.go
tkong-redhat: I used get here because here also means the read legacy-token failed to.

pkg/maintenance/alertmanagerMaintenance.go
tkong-redhat: I used "get" here because here also means the read legacy-token failed to.

PR #445, opened 2024-07-08
PDB HealthCheck Implementation

shitaljante:
### What type of PR is this?
feature

PDB HealthCheck Implementation

[OSD-22456](https://issues.redhat.com//browse/OSD-22456)

Screenshot 2024-07-10 at 9 03 34 AM


pkg/dvo/builder.go
```diff
@@ -0,0 +1,56 @@
...
+func (dcb *dvoClientBuilder) New(c client.Client) (DvoClient, error) {
+
+	// Get the service URL for the deployment-validation-operator-metrics service
+	svcURL, err := metrics.NetworkTarget(c, "openshift-deployment-validation-operator", "deployment-validation-operator-metrics", "8383")
+	if err != nil {
+		return nil, err
+	}
+
+	// Print the service URL for debugging purposes
+	fmt.Printf("********%s", svcURL)
```
tkong-redhat, line 36: Will we keep this in future? Or it will be removed before merge

shitaljante: yeah I will make sure to remove the testing loggers before merge. Thank you

hack/olm-registry/olm-artifacts-template.fedramp.yaml
```diff
@@ -136,6 +136,20 @@ objects:
             - get
             - list
             - watch
+        - apiVersion: rbac.authorization.k8s.io/v1
+          kind: Role
```
ravitri, line 140: Adding of these resources is actually subject to if DVO is deployed in Fedramp env as well or not. Am checking internally if it's deployed or not.

pkg/dvo/builder.go
```diff
@@ -0,0 +1,54 @@
...
+	// Get the service URL for the deployment-validation-operator-metrics service
+	svcURL, err := metrics.NetworkTarget(c, "openshift-deployment-validation-operator", "deployment-validation-operator-metrics", "8383")
+	if err != nil {
+		return nil, err
+	}
+
+	// Fetch the cluster AccessToken
+	accessToken, err := util.GetAccessToken(c)
+	if err != nil {
+		fmt.Printf("err accestoken%s", err)
```
ravitri, line 38: Do we need this line since the line below already logs the error?

pkg/dvo/client.go
```diff
@@ -0,0 +1,76 @@
...
+	return transport.RoundTrip(req)
+}
+
+func (c *dvoClient) GetMetrics() ([]byte, error) {
+	// Construct the URL for the metrics API
+	metricsURL := "http://" + c.dvoBaseUrl + METRICS_API_PATH
+
+	req, err := http.NewRequest("GET", metricsURL, nil)
+	if err != nil {
+		return nil, fmt.Errorf("could not query prometheus: %s", err)
```
ravitri, line 62:
```suggestion
		return nil, fmt.Errorf("could not query DVO metrics endpoint: %s", err)
```

pkg/upgraders/healthcheck_pdb.go
```diff
@@ -0,0 +1,84 @@
...
+		if !strings.HasPrefix(pdb.Namespace, "openshift-*") {
+			if pdb.Spec.MaxUnavailable != nil && pdb.Spec.MaxUnavailable.IntVal == 0 {
+				//BAD pdb
+				return fmt.Errorf("found a PodDisruptionBudget with MaxUnavailable set to 0")
+			} else if pdb.Status.CurrentHealthy < pdb.Status.DesiredHealthy {
+				//BAD pdb
+				return fmt.Errorf("found a PodDisruptionBudget with CurrentHealthy less than DesiredHealthy")
+			}
+		}
+	}
```
ravitri, line 52: Some points to note as discussed:
-  Based on https://github.com/openshift/managed-cluster-config/blob/master/deploy/managed-upgrade-operator-config/10-managed-upgrade-operator-configmap.yaml#L78 we do have few `openshift-*` namespace that we want to treat as user namespaces which might have workloads installed by users and are not platform workloads. We can include them as an exception to `openshift-*` pattern.
- The `else` block could have more implications since it could refer to a transient state which will resolve in some time (or may not resolve in some time). We can start with the static checks in the `if` block only. 
- It would also help to include returning the `reason` label from this function if we are considering different scenarios for PDB failure.

pkg/upgraders/healthcheckstep.go
```diff
@@ -58,6 +58,15 @@ func (c *clusterUpgrader) PreUpgradeHealthCheck(ctx context.Context, logger logr
 		logger.Info(fmt.Sprintf("upgrade delayed due to there are unschedulable taints on nodes: %s", err))
 		healthCheckFailed = append(healthCheckFailed, "NodeUnschedulableTaintHealthcheckFailed")
 	}
+	
+	// add error handling and servicelog send
+	ok, err = HealthCheckPDB(c.metrics, c.client, c.dvo)
+	if err != nil || !ok {
+		logger.Info(fmt.Sprintf("upgrade delayed due PDB %s", err))
+		//errResult := c.notifier.Notify(notifier.MuoStateHealthCheck)
```
ravitri, line 66:
```suggestion
```

pkg/upgraders/healthcheck_pdb.go
shitaljante: Updated changes,Thank you.

PR #447, opened 2024-07-12
OSD-23975 add version and state in the health-check metrics 

tkong-redhat:
Add version and state in the health-check metrics so the health_check could be exposed to RHOBS


ravitri:
/hold

@tkong-redhat - Can we include the changes related to https://github.com/openshift/managed-upgrade-operator/pull/445 as well? The PR should merge in a while

pkg/upgraders/healthcheckstep.go
```diff
@@ -119,5 +122,28 @@ func (c *clusterUpgrader) PostUpgradeHealthCheck(ctx context.Context, logger log
 	if err != nil || !ok {
 		return false, err
 	}
+
 	return true, nil
 }
+
+func getCurrentVersion(ug *upgradev1alpha1.UpgradeConfig) string {
```
ravitri, line 129: Can we use ClusterVersion version here?

tkong-redhat: Nice suggestion. Let me have a look

ravitri:
/lgtm
/approve

Thanks a lot Tony!

PR #452, opened 2024-07-18
OSD-24706: Adding test file for ocmprovider's config

TheUndeadKing:
### What type of PR is this?
unit test

### JIRA?
[OSD-24722](https://issues.redhat.com/browse/OSD-24722)

Adding unit test to `pkg/ocmprovider/config.go`


shitaljante:
Can we add a test to cover the scenario- 
`		return fmt.Errorf("OCM Base URL is not a parseable URL")
`
So the file coverage will be 100%.

PR #456, opened 2024-07-24
OSD-18463: MUO to report upgrade didn't started due to edge removal

TheUndeadKing:
### What type of PR is this?
feature

If edge has been removed before upgrade for some reason. The MUO silently fails because it knows the edge has been removed but OCM doesn’t the edge has been removal. This PR wants to fulfill the following requirement. [**Requirement:** MUO should notify the OCM that the upgrade fails because of edge has been removed. `MUO STATE: Failed (OR) Cancelled & the reason should be "Edge has been removed"`]

[OSD-18463](https://issues.redhat.com/browse/OSD-18463)


ravitri:
/close

As discussed, there's a separate plan to handle this for long term so we will get back to it in future. For now, am closing this PR. Thanks a lot Anwar ! :+1:

PR #457, opened 2024-07-25
Make no worker machinepool PHC as non-failing

chamalabey:
### What type of PR is this?
_feature_

For the no worker machinepool PHC, we do not want to fail the PHC. But we would still like to continue notifying the customer that the upgrade will be delayed since there's no worker machinepool. If there's no other PHC failure and there's no worker machinepool, the PHC should not fail.

_Fixes # [OSD-24705](https://issues.redhat.com/browse/OSD-24705)

### DIFF
--- a/pkg/upgraders/healthcheckstep.go
+++ b/pkg/upgraders/healthcheckstep.go
@@ -63,9 +63,12 @@ func (c *clusterUpgrader) PreUpgradeHealthCheck(ctx context.Context, logger logr
 		if c.upgradeConfig.Spec.CapacityReservation {
 			ok, err := c.scaler.CanScale(c.client, logger)
 			if !ok || err != nil {
-				capacityReservationFailed = true
+				// HealthCheck should not fail in this senario. Have to inform the customer about the upgrade delay.
 				c.metrics.UpdateMetricHealthcheckFailed(c.upgradeConfig.Name, metrics.DefaultWorkerMachinepoolNotFound, version, state)
-				healthCheckFailed = append(healthCheckFailed, "CapacityReservationHealthcheckFailed")
+				err = c.notifier.Notify(notifier.MuoStateDelayedWorkerMachinepoolNotFound)
+				if err != nil {
+					return false, err
+				}
 			} else {
 				logger.Info("Prehealth check for CapacityReservation passed")
 				c.metrics.UpdateMetricHealthcheckSucceeded(c.upgradeConfig.Name, metrics.DefaultWorkerMachinepoolNotFound, version, state)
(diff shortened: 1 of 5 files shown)


pkg/upgraders/healthcheckstep.go
```diff
@@ -40,8 +40,12 @@ func (c *clusterUpgrader) PreUpgradeHealthCheck(ctx context.Context, logger logr
 	if c.upgradeConfig.Spec.CapacityReservation {
 		ok, err := c.scaler.CanScale(c.client, logger)
 		if !ok || err != nil {
+			// HealthCheck should not fail in this senario. Have to inform the customer about the upgrade delay.
 			c.metrics.UpdateMetricHealthcheckFailed(c.upgradeConfig.Name, metrics.DefaultWorkerMachinepoolNotFound)
-			healthCheckFailed = append(healthCheckFailed, "CapacityReservationHealthcheckFailed")
+			err = c.notifier.Notify(notifier.MuoStateDelayed)
```
tkong-redhat, line 45: This will create a SL then set the metric. If other PHC failed later, SL will not be sent out because SL sent metric has been set.

pkg/upgraders/healthcheckstep.go
```diff
@@ -40,8 +40,12 @@ func (c *clusterUpgrader) PreUpgradeHealthCheck(ctx context.Context, logger logr
...
+			if err != nil {
+				return false, err
```
shitaljante, line 47: If it fails to Notify, will it block the next pre-health check?

pkg/upgraders/healthcheckstep.go
shitaljante: I think SL sent metrics is based on MUO state which will be different in PHC . 
`s.metrics.UpdateMetricNotificationEventSent(uc.Name, string(state), uc.Spec.Desired.Version)`

But good to verify this once.

pkg/upgraders/healthcheckstep.go
chamalabey: NotificationEventSent metric set in the end just before nil error return, hence if there is an error it will return before the metrics gets update.
https://github.com/openshift/managed-upgrade-operator/blob/master/pkg/eventmanager/eventmanager.go#L159

pkg/upgraders/healthcheckstep.go
chamalabey: I guess this might effect the UpgradeDelayCheck notification here https://github.com/openshift/managed-upgrade-operator/blob/master/pkg/upgraders/delaystep.go#L38 
Any suggestions on how to notify the user without other impact ?

chamalabey:
Rebased

pkg/upgraders/healthcheckstep.go
tkong-redhat: I would recommend to add an extra flag which indicate if the default worker pool check failed then adding the information accordingly.

chamalabey:
Closing this due to replacement with https://github.com/openshift/managed-upgrade-operator/pull/466

PR #461, opened 2024-08-01
revert the changes in fedramp rbac which causes problem

tkong-redhat:
### What type of PR is this?
_(bug)_

Revert the RBAC added for DVO PDB checking to unblock the Fedramp team

_Fixes #_


ravitri:
/lgtm
/approve

Thanks a lot Tony :+1:

PR #476, opened 2024-10-04
OSD-25179 - Avoid deadlock when lockfile is presetn

Tafhim:
### What type of PR is this?
bug

MUO can sometimes get stuck in a deadlock if a pre-existing lockfile has ownerReference mismatch. This PR pre-emtively removes the lock file to avoid such scenario.

[OSD-25179](https://issues.redhat.com/browse/OSD-25179)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -30,6 +30,8 @@ require (
 	github.com/openshift/machine-config-operator v0.0.1-0.20230330142923-2832f049b3f4
 	github.com/openshift/operator-custom-metrics v0.5.1
 	github.com/openshift/osde2e-common v0.0.0-20240531074950-36a7055798ae
+	// Note: Please revisit the stop-gap in ./main.go > cleanupLockForLeader when updating operator-lib module.
+	// If the issue has been fixed upstream, the stop-gap should be removed.
 	github.com/operator-framework/operator-lib v0.12.0
 	github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.71.0
 	github.com/prometheus-operator/prometheus-operator/pkg/client v0.71.0
--- a/main.go
+++ b/main.go
@@ -97,7 +97,11 @@ var (
 	scheme                  = apiruntime.NewScheme()
 	setupLog                = ctrl.Log.WithName("setup")
 )
-var log = logf.Log.WithName("cmd")
+
+var (
+	log                  = logf.Log.WithName("cmd")
+	muoLockConfigMapName = "managed-upgrade-operator-lock"
+)
 
 func init() {
 	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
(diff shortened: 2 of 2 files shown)


go.mod
```diff
@@ -30,6 +30,8 @@ require (
 	github.com/openshift/machine-config-operator v0.0.1-0.20230330142923-2832f049b3f4
 	github.com/openshift/operator-custom-metrics v0.5.1
 	github.com/openshift/osde2e-common v0.0.0-20240531074950-36a7055798ae
+	// Note: Please revisit the stop-gap in ./main.go > cleanupLockForLeader when updating operator-lib module.
+	// If the issue has been fixed upstream, the stop-gap should be removed.
```
rendhalver, line 34: Thanks Tafhim!
These are great!

ravitri:
/hold

Thanks for reviews so far! Adding temporary hold on this PR till it's further reviewed. I did have some comments

go.mod
ravitri: @Tafhim - I think it's better to have such notes added in markdown or the code pkg files for reference. However, if we want to ensure that a go.mod version is tied to version due to a known issue etc and we don't accidentally update it then it might be better to have it commented with the upstream issue/bug as follows under the `replace` directive. Example:
```
replace (
    github.com/operator-framework/operator-lib v0.12.0 // Refer upstream issue <link>
)
```
It would also be fine to have a "maintenance" section added in this repo's docs - https://github.com/openshift/managed-upgrade-operator/blob/master/docs/development.md 

We can have a separate maintenance.md file covering overall process and highlight this dependency requirement/reason there.

main.go
```diff
@@ -117,6 +121,70 @@ func printVersion() {
 	log.Info(fmt.Sprintf("Version of operator-sdk: v%v", version.SDKVersion))
 }
 
+// This function is used to pre-emptively avoid a deadlock
+// Note: This is a stop-gap solution until a fix has been implemented
+// upstream in operator-framework/operator-lib. This function and it's
+// usage should be removed once a fix is implemented upstream.
```
ravitri, line 127: Instead of `Note`, can it be replaced as a `TODO` point?

go.mod
rendhalver: I suggested putting it here so it would be noticed if the version was updated.
As far as I am aware we don't have the link for the upstream issue yet so a more generic note was suggested.
Putting this somewhere it's not going to be noticed may mean the patched release will happen without anyone noticing and we will have  two services updating the same resource which may cause issues.

ravitri:
/close

Based on discussion with @Tafhim, we will go ahead with PR https://github.com/openshift/managed-upgrade-operator/pull/483 changes for now and observe if it helps ahead. We believe that the operator-lib bump would help address the issue we were facing (atleast partially). Incase it still continues to impact us, we will open a new PR based on this PR's work (might need to align the solution a bit based on what exact problem we are trying to address).

However, this PR has helped us revise some of the implementation details for the lockfile which was great and will serve good reference for future too. Thanks for your help and work on this Tafhim :+1:  

Closing this PR for now.

PR #494, opened 2025-01-23
OSD-26790: Adding metric of SL failure

TheUndeadKing:
### What type of PR is this?
_(bug)_

When the user gets banned (OR) PS has been rotated in a middle of the upgrade, the upgrade will stall. We're creating a metric if MUO unable to notify it's state to the OCM. This metric will help us to track via PD, so that we can unblock the upgrade.

_Fixes #_ [OSD-26790](https://issues.redhat.com//browse/OSD-26790)

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -8,7 +8,7 @@ COPY . .
 RUN make go-build
 
 ####
-FROM registry.access.redhat.com/ubi8/ubi-minimal:8.10-1154
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.10-1179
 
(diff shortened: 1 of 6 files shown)


a7vicky:
Thanks for the PR @TheUndeadKing 
I have a couple of inputs:
1. Thinking of setting the value to `1` because metrics naming has a suffix `failed`
2. Can we Include a state label on the metrics
3. Please squash the commit and make sure PR checks are passing

a7vicky:
Please include the logic to reset the metric if MUO is successfully able to send Servicelog

Issue #496, opened 2025-02-24
"Observed a panic in reconciler: odd number of arguments passed as key-value pairs for logging"


newera-begins:
While testing upgrade Config in a local mode it's failed to parse an `upgradeAt` with the below panic error.

```
ts=2025-02-17T11:31:36.694027421Z level=info msg="Starting workers" controller=node controllerGroup= controllerKind=Node workercount=1
ts=2025-02-17T11:31:37.096761184Z level=info logger=upgrade-config-manager msg="Starting the upgradeConfigManager"
ts=2025-02-17T11:31:38.093767169Z level=info logger=controller_upgradeconfig msg="Current cluster status" Request.Namespace=openshift-managed-upgrade-operator Request.Name=managed-upgrade-config status=New
ts=2025-02-17T11:31:38.093982303Z level=dpanic logger=scheduler msg="odd number of arguments passed as key-value pairs for logging" ignoredkey=2025-02-017T11:20:00Z stacktrace="github.com/openshift/managed-upgrade-operator/pkg/scheduler.(*scheduler).IsReadyToUpgrade\n\t/workdir/pkg/scheduler/scheduler.go:40\ngithub.com/openshift/managed-upgrade-operator/controllers/upgradeconfig.(*ReconcileUpgradeConfig).Reconcile\n\t/workdir/controllers/upgradeconfig/upgradeconfig_controller.go:165\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile\n\tpkg/mod/sigs.k8s.io/controller-runtime@v0.17.0/pkg/internal/controller/controller.go:119\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\tpkg/mod/sigs.k8s.io/controller-runtime@v0.17.0/pkg/internal/controller/controller.go:316\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\tpkg/mod/sigs.k8s.io/controller-runtime@v0.17.0/pkg/internal/controller/controller.go:266\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\tpkg/mod/sigs.k8s.io/controller-runtime@v0.17.0/pkg/internal/controller/controller.go:227"
[...]

PR #497, opened 2025-02-24
issue:496  Fixed odd number of arguments passed as key-value pairs

newera-begins:
### What type of PR is this?
_Bug_

_Fixes #_
https://github.com/openshift/managed-upgrade-operator/issues/496

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -15,6 +15,7 @@ var (
 )
 
 // Scheduler is an interface that enables implementations of type Scheduler
+//
 //go:generate mockgen -destination=mocks/mockScheduler.go -package=mocks github.com/openshift/managed-upgrade-operator/pkg/scheduler Scheduler
 type Scheduler interface {
 	IsReadyToUpgrade(*upgradev1alpha1.UpgradeConfig, time.Duration) SchedulerResult
@@ -37,7 +38,7 @@ type SchedulerResult struct {
 func (s *scheduler) IsReadyToUpgrade(upgradeConfig *upgradev1alpha1.UpgradeConfig, timeOut time.Duration) SchedulerResult {
 	upgradeTime, err := time.Parse(time.RFC3339, upgradeConfig.Spec.UpgradeAt)
 	if err != nil {
-		logger.Error(err, "failed to parse spec.upgradeAt", upgradeConfig.Spec.UpgradeAt)
+		logger.Error(err, "failed to parse spec.upgradeAt", "upgradeAt", upgradeConfig.Spec.UpgradeAt)
 		return SchedulerResult{IsReady: false, IsBreached: false, TimeUntilUpgrade: 0}
 	}
 	now := time.Now()


ravitri:
@newera-begins - Yup, it's a valid bug and fix. Thanks a lot ! :+1: 

Just to elaborate on this further (would be better to update PR description too)
- This issue is only when `upgradeConfig.Spec.UpgradeAt` is incorrect and is not specific to local mode only.
- There are few reasons it's missed so far:
  - `OCM` mode is default and that would always populate the correct time format.
  - LOCAL mode is comparatively less used but regardless, the panic should be fixed.
  - These two lines do not have a unit test coverage as well so that's another reason why it wasn't caught so far:
    ```go
	  if err != nil {
		  logger.Error(err, "failed to parse spec.upgradeAt", "upgradeAt", upgradeConfig.Spec.UpgradeAt)
		  return SchedulerResult{IsReady: false, IsBreached: false, TimeUntilUpgrade: 0}
	  }
  ```
Can you add a small unit test to handle this scenario as well? Mainly, `IsReadyToUpgrade` method should return the SchedulerResult accordingly if the `UpgradeAt` is malformed.

newera-begins:
> @newera-begins - Yup, it's a valid bug and fix. Thanks a lot ! 👍
> 
> Just to elaborate on this further (would be better to update PR description too)
> 
> * This issue is only when `upgradeConfig.Spec.UpgradeAt` is incorrect and is not specific to local mode only.
> * There are few reasons it's missed so far:
>   
>   * `OCM` mode is default and that would always populate the correct time format.
>   * LOCAL mode is comparatively less used but regardless, the panic should be fixed.
>   * These two lines do not have a unit test coverage as well so that's another reason why it wasn't caught so far:
>     ```go
>       if err != nil {
>     	  logger.Error(err, "failed to parse spec.upgradeAt", "upgradeAt", upgradeConfig.Spec.UpgradeAt)
>     	  return SchedulerResult{IsReady: false, IsBreached: false, TimeUntilUpgrade: 0}
>       }
>     ```
> 
> Can you add a small unit test to handle this scenario as well? Mainly, `IsReadyToUpgrade` method should return the SchedulerResult accordingly if the `UpgradeAt` is malformed.

@ravitri sure I will write this. Thanks for the confirmation.

ravitri:
@newera-begins - Can you rebase your PR? We need to have this fix on top of https://github.com/openshift/managed-upgrade-operator/pull/498 that's merged. Likely to pass the validate test then

go.mod
```diff
@@ -108,6 +113,7 @@ require (
 	github.com/skratchdot/open-golang v0.0.0-20200116055534-eef842397966 // indirect
 	github.com/spf13/cobra v1.8.0 // indirect
 	github.com/spf13/pflag v1.0.5 // indirect
+	github.com/stretchr/testify v1.10.0
```
ravitri, line 116: @newera-begins - Sorry, just noticed it now. It would be better to maintain the testing practices same as https://github.com/openshift/managed-upgrade-operator/blob/master/docs/testing.md which is applicable for all the packages in the repo. Although it will work, we don't want to introduce separate testing practices via different tools at this point. Can you rework the test accordingly? Even a single test that's simple will suffice for now. So you can rebase your PR branch, rework the test and then push the changes.

newera-begins:
@ravitri please review and let me know in case of any modifications. Thanks

a7vicky (commented):
Looks good to me.
@ravitri is it okay to remove hold ?

a7vicky:
tide/merge-method-squash

PR #508, opened 2025-04-23
Osd 28839 mc should fire alert when node drain takes long time

tkong-redhat:
### What type of PR is this?
This is a bug fix

The PR make sure nodekeeper controller is still monitoring the node drain process and set node drain failed metric when node drain takes long time.
[Osd-28839](https://issues.redhat.com//browse/Osd-28839)

Because nodedrain strategy is disabled on MC currently. There is no node drain failed metric is set when node drain takes long time. SRE is not able to know the node drain is stuck for a long time on the MC which lead to long upgrade time on the MC.
The PR make sure there is still default drain strategy is setup to monitoring the node drain process and set the metric when node drain takes long time

### DIFF
--- a/controllers/nodekeeper/nodekeeper_controller.go
+++ b/controllers/nodekeeper/nodekeeper_controller.go
@@ -145,7 +145,7 @@ func (r *ReconcileNodeKeeper) SetupWithManager(mgr ctrl.Manager) error {
 		Complete(r)
 }
 
-// SetupWithManager sets up the controller with the Manager.
+// Check the NodeDrainResult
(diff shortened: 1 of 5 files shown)


controllers/nodekeeper/nodekeeper_controller.go
```diff
@@ -148,3 +144,24 @@ func (r *ReconcileNodeKeeper) SetupWithManager(mgr ctrl.Manager) error {
 		WithEventFilter(IgnoreMasterPredicate()).
 		Complete(r)
 }
+
+// SetupWithManager sets up the controller with the Manager.
```
a7vicky, line 148:
```suggestion
```

tkong-redhat: Updated

PR #514, opened 2025-05-16
demonstrate how to check if a node is stuck draining

deads2k:
Adds a StuckNode condition on the ManageUpgradeOperator.  This is slightly more than simply: has it taken a while. It does the following

1. have I observed the same node draining for more than a period of time
2. if I have, report the following
   1. report all still running pods with very long termination grace period seconds
   2. report all still running pods protected by PDBs that are preventing eviction and indication which PDBs are preventing eviction.

We have choices about whether to report pods with long termination grace period second under StuckNode or report them under a different condition since we know they will eventually be cleaned up.

Assuming we agree on the utility of change, things left to do:
1. unit tests.  Probably easiest to drive based on must-gather loading using https://github.com/openshift/library-go/tree/master/pkg/manifestclienttest so that any bugs can easily have test cases added for them
2. wiring into the running logic based on a 10 minute clock.  It's currently just a separate POC.
3. education on server-side apply, plus adding a library similar to the library-go diff logic to allow a dynamic client to serialize and set this status.

### DIFF
--- a/api/v1alpha1/upgradeconfig_types.go
+++ b/api/v1alpha1/upgradeconfig_types.go
@@ -53,6 +53,13 @@ type UpgradeConfigSpec struct {
 
 // UpgradeConfigStatus defines the observed state of UpgradeConfig
 type UpgradeConfigStatus struct {
+	// Represents the observations of a foo's current state.
+	// This is the kube standard conditions stanza.
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=type
+	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
 
 	// This record history of every upgrade
 	// +kubebuilder:validation:Optional
--- a/controllers/stucknodecondition/stuck_node_checker.go
+++ b/controllers/stucknodecondition/stuck_node_checker.go
@@ -1,0 +1,341 @@
+package upgradeconfig
+
+import (
+	"context"
+	"fmt"
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	policyv1 "k8s.io/api/policy/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/utils/clock"
+	"sort"
+	"strings"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+var (
+	log = logf.Log.WithName("controller_stucknode")
+)
+
+// blank assignment to verify that StuckNode implements reconcile.Reconciler
+var _ reconcile.Reconciler = &StuckNode{}
+
+// StuckNode reconciles the StuckNode condition.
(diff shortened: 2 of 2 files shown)


controllers/stucknodecondition/stuck_node_checker.go
```diff
@@ -0,0 +1,341 @@
...
+
+	drainingNodes := []corev1.Node{}
+	drainingNodesByName := map[string]corev1.Node{}
+	for _, node := range allNodes.Items {
+		// draining nodes are unschedulable
+		if !node.Spec.Unschedulable {
+			continue
+		}
+		// the MCO uses this annotation with a suffix of drain to indicate draining.
+		if !strings.HasSuffix(node.Annotations["machineconfiguration.openshift.io/desiredDrain"], "drain") {
```
wking, line 71: Checking in case this is too new for our management clusters:

```console
$ for Y in 10 11; do git --no-pager grep 'machineconfiguration.openshift.io/desiredDrain' "origin/release-4.${Y}"; done
origin/release-4.11:pkg/daemon/constants/constants.go:  DesiredDrainerAnnotationKey = "machineconfiguration.openshift.io/desiredDrain"
```

So all 4.(y>=11) management clusters should have an MCO that sets this annotation, and our management clusters are all 4.15 or 4.16 or newer or something.  So should be fine as you have it, but commenting in case other folks also wonder about this.

controllers/stucknodecondition/stuck_node_checker.go
```diff
@@ -0,0 +1,341 @@
...
+
+// checkIfPodIsEvictable reasons why the pod is not evictable
+func checkIfPodIsEvictable(ctx context.Context, kubeClient kubernetes.Interface, pod corev1.Pod) []error {
+	// terminal pods are not subject to PDBs
+	if canIgnorePDB(pod) {
+		return nil
+	}
+
+	// TODO option on caching, might not matter
+	allPDBsInNamespace, err := kubeClient.PolicyV1().PodDisruptionBudgets(pod.Namespace).List(ctx, metav1.ListOptions{})
```
wking, line 220: There's likely some overlap between this kind of check and [what the managed-upgrade operator is already doing][1], but I'll leave it up to actual MUO maintainers to decide if/how to integrate this with their existing code.

[1]: https://github.com/openshift/managed-upgrade-operator/blob/14278e8ffc0d3e54f326dbc20501af08a9491fbb/pkg/drain/strategy.go#L101-L122

api/v1alpha1/upgradeconfig_types.go
```diff
@@ -53,6 +53,13 @@ type UpgradeConfigSpec struct {
 
 // UpgradeConfigStatus defines the observed state of UpgradeConfig
 type UpgradeConfigStatus struct {
+	// Represents the observations of a foo's current state.
+	// This is the kube standard conditions stanza.
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=type
+	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
```
a7vicky, line 62: Thanks for the PR @deads2k.

Currently one of the PR check if failing because CRD is not generated after this change. We are using `controller-gen` for generating CRD: 
`make container-generate`

controllers/stucknodecondition/stuck_node_checker.go
a7vicky: +1 We have something similar as part of [nodekeeper controller](https://github.com/openshift/managed-upgrade-operator/blob/master/controllers/nodekeeper/nodekeeper_controller.go#L125) logic. I am thinking if we can merge these.

Also, we did disable these drain strategy for managenent cluster to identify if there is any issue with the workload which we are running.

PR #518, opened 2025-06-06
SREP-368 add note for MacOS Developer

charlesgong:
### What type of PR is this?
documentation

add note for MacOS developer

SREP-368 

_Fixes #_


docs/development.md
```diff
@@ -113,6 +113,8 @@ To run lint, test and build in `app-sre/boilerplate` container, call `boilerplat
 ```shell
 $ boilerplate/_lib/container-make
 ```
+### Note for macOS Developers:
+When building on macOS, the boilerplate container-based build approach `make container-build` is not supported due to limitations with running Podman-in-Podman in a cross-platform environment. Instead, macOS users should use `make go-build` with the architect flag to build x86-compatible binaries example: `CGO_ENABLED=0 GOOS=linux GOARCH=amd64 make go-build`
```
ravitri, line 117:
```suggestion
When building on macOS, the boilerplate container-based build approach `make container-build` is not supported due to limitations with running Podman-in-Podman in a cross-platform environment. Instead, macOS users should use `make go-build` with the architecture flag to build x86-compatible binaries example: `CGO_ENABLED=0 GOOS=linux GOARCH=amd64 make go-build`
```

joshbranham: Can you embed this context in the make targets or some script versus requiring uses to take manual action?

PR #525, opened 2025-07-01
SREP-669 - Add ephemeral-storage request

Tafhim:
### What type of PR is this?
bug

This PR is needed to address a bug in MUO. The bug occurs due to a lack of request set on ephemeral storage. Multiple pod evictions happens due to not having the request set.

[SREP-669](https://issues.redhat.com//browse/SREP-669)

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -48,6 +48,7 @@ spec:
             requests:
               cpu: 20m
               memory: 100Mi
+              ephemeral-storage: 30Ki
             limits:
               cpu: 50m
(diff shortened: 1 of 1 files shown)


deploy/operator.yaml
```diff
@@ -48,6 +48,7 @@ spec:
             requests:
               cpu: 20m
               memory: 100Mi
+              ephemeral-storage: 30Ki"
```
ravitri, line 51: Can you double check the quotes here? Is it a typo?

PR #552, opened 2025-09-03
Feat(SREP-1600): handle EUS upgrade channel switches

typeid:
### What type of PR is this?
feature

For the EUS (Extended Update Support) channel, which is being introduced with https://issues.redhat.com/browse/XCMSTRAT-1119, we need to consume the right channels.

EUS channels follow a specific versioning pattern where they only exist for even minor versions (e.g., 4.16, 4.18, 4.20). When MUO attempts to upgrade to an odd minor version via an EUS channel (e.g., eus-4.17), the channel doesn't exist, which would cause upgrade failures.

This PR modifies the `inferUpgradeChannelFromChannelGroup` function to automatically round up odd minor versions to the next even number for EUS channel groups. For example:
 - `eus` + version `4.17.1` now produces `eus-4.18` instead of `eus-4.17`
 - `eus` + version `4.18.1` continues to produce `eus-4.18` (no change for even versions)
 - Non-EUS channels remain unaffected

Fixes https://issues.redhat.com/browse/SREP-1600 


andreadecorte:
lgtm

PR #600, opened 2026-04-17
chore: update all OLM template SSSs to Upsert

Ajpantuso:
### What type of PR is this?
_chore_

Updates all SelectorSyncSets in the OLM template to `resourceApplyMode: Upsert`.
This will allow a safe transition to PKO when the old SelectorSyncSets are deleted.
When in `Upsert` mode Hive orphans the resources applied to clusters matching the SelectorSyncSet rather than deleting them when the SelectorSyncSet is deleted.


dustman9000 (commented):
/lgtm

PR #607, opened 2026-05-04
[SREP-3999] DVO NS isn't present in MC and SC. Which impact the install

TheUndeadKing:
### What type of PR is this?
_(refactor)_

The DVO NS isn't present in some cluster(Mc and sc), which blocks the unpacking of clusterpackage resource.

_Fixes #_


TheUndeadKing:
/close
We are moving DVO inside SSS instead of phase

PR #694, opened 2026-08-11
Remove accesstoken from DVO endpoint access

chamalabey:
### What type of PR is this?
_bug_

Remove the Authorization header from dvoRoundTripper entirely (DVO /metrics is unauthenticated) so the pull-secret is never attached

_Fixes #_[ROSAENG-61337](https://redhat.atlassian.net/browse/ROSAENG-61337)

[ROSAENG-61337]: https://redhat.atlassian.net/browse/ROSAENG-61337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added support for overriding the DVO service URL with `DVO_SVC_URL`.
  * Added local port forwarding for DVO metrics on port 53083.
  * Added configurable retry limits and execution deadlines for end-to-end test jobs.

* **Bug Fixes**
  * Improved DVO connectivity and transport configuration.
  * Preserved response bodies for unsuccessful requests.
  * Improved handling and retrying of failed or timed-out test executions.

* **Documentation**
  * Updated local development instructions for running with internal cluster services.

* **Tests**
  * Added coverage for metrics retrieval, connectivity, proxy support, TLS timeouts, and response handling.


TheUndeadKing:
You'd completely removed the accesstoken mechanism, this might affect the local run[1] because it connects to route not svc.

[1] https://github.com/openshift/managed-upgrade-operator/blob/master/pkg/metrics/metrics.go

ref: https://github.com/openshift/ops-sop/blob/master/project_guides/managed-upgrade-operator/hack/muo-testing.sh

I think we don't need the local run test anymore. because of PKO migration, it makes our life easy around to deploy and test our operator as image rather than running it locally.

chamalabey:
> You'd completely removed the accesstoken mechanism, this might affect the local run[1] because it connects to route not svc.
> 
> [1] https://github.com/openshift/managed-upgrade-operator/blob/master/pkg/metrics/metrics.go
> 
> ref: https://github.com/openshift/ops-sop/blob/master/project_guides/managed-upgrade-operator/hack/muo-testing.sh
> 
> I think we don't need the local run test anymore. because of PKO migration, it makes our life easy around to deploy and test our operator as image rather than running it locally.

@TheUndeadKing Those files have not been updated for two years now. We might want to remove those. 
There is no need to have authorization if that is not required with in the cluster hence I removed it. 
Yes we have to use image to test the operator locally. 
Please let me know your thoughts?

pkg/dvo/builder.go
```diff
@@ -28,19 +28,21 @@ func (dcb *dvoClientBuilder) New(c client.Client) (DvoClient, error) {
 
 	// Get the service URL for the deployment-validation-operator-metrics service
 	svcURL, err := metrics.NetworkTarget(c, "openshift-deployment-validation-operator", "deployment-validation-operator-metrics", "http-metrics")
-	if err != nil {
-		return nil, err
+
+	// For local run of MUO via `make run`, it's expected that DVO_SVC_URL is set, after port-forwarding DVO metrics service
```
Tafhim: I don't know how useful it is to put this knowledge here. It seems it belong in a guide or testing documentation. I don't think testers would look for this information here. Maybe we should move these instructions in the testing doc.

chamalabey: I have moved the instructions to the docs. Thank you Tafhim

pkg/dvo/client.go
```diff
@@ -38,26 +31,15 @@ type dvoClient struct {
...
-	authorization util.AccessToken
-}
-
-func (drt *dvoRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
-	authVal := fmt.Sprintf("AccessToken %s:%s", drt.authorization.ClusterId, drt.authorization.PullSecret)
-	req.Header.Add("Authorization", authVal)
-	transport := http.Transport{
-		// Configure proxy support for Routes mode (when DVO is accessed externally)
-		// Respects HTTP_PROXY, HTTPS_PROXY, and NO_PROXY environment variables
+func newDvoTransport() *http.Transport {
```
tkong-redhat: If we only have one Transport. Then the `new` prefix doesn't make sense.
Please follow the naming best practice to make it meaningful.

chamalabey: I have renamed it. Thank you Tony

PR #705, opened 2026-09-09
fix(#698): update channel on image-based upgrades

### LINKED ISSUE #698: Channel is not updated when both spec.desired.image and spec.desired.channel are set in UpgradeConfig
### Summary

When an `UpgradeConfig` has `spec.desired.image` populated alongside `spec.desired.channel`
and `spec.desired.version`, the operator performs the version upgrade but never updates the
cluster's channel to match `spec.desired.channel`. This means that after a cross-minor
upgrade (e.g. stable-4.20 -> stable-4.21), the cluster's channel stays on the old value and
has to be changed manually afterwards via `oc adm upgrade channel`.

### Steps to reproduce

1. Cluster is on channel `stable-4.20`, running version `4.20.27`.
2. Create/update an `UpgradeConfig` with all three desired fields populated, e.g.:
   ```yaml
   spec:
     desired:
       channel: stable-4.21
       image: quay.io/openshift-release-dev/ocp-release@sha256:<digest>
       version: 4.21.27
3. The upgrade proceeds and the cluster is successfully updated to version 4.21.27.
4. The cluster's channel remains stable-4.20 instead of being updated to stable-4.21,
even though channel: stable-4.21 was explicitly set in the UpgradeConfig.

### Expected behavior

When spec.desired.channel is set and differs from the cluster's current channel, the
operator should update the cluster's channel to match, regardless of whether the upgrade
is being driven via spec.desired.image or via spec.desired.channel + spec.desired.version.
Today this only appears to happen for the latter case, not when image is also set.

### Impact

Admins/automation that populate image (e.g. to pin an exact release) together with
channel for cross-minor upgrades end up with a cluster left on the wrong channel after
the upgrade completes, requiring a manual follow-up step to correct it. This is easy to
miss and can go unnoticed until the next upgrade cycle.

pazpuwi:
## Description

Fixes #698

When an `UpgradeConfig` has `spec.desired.image` populated alongside `spec.desired.channel` and `spec.desired.version`, the operator performs the version upgrade but never updates the cluster's channel to match `spec.desired.channel`. After a cross-minor image-based upgrade (e.g. `stable-4.20` -> `stable-4.21`), the cluster is left on the old channel and must be corrected manually via `oc adm upgrade channel`.

The root cause is that when `spec.desired.image` is set, `checkUpgradeSource` routes the upgrade through `runUpgradeWithImage`, which only patched `spec.desiredUpdate.image` and never touched the channel. Channel syncing only happened in `runUpgradeWithChannelVersion`.

## Change

`runUpgradeWithImage` now syncs the cluster channel before patching the desired image when `spec.desired.channel` is set and differs from the current `cv.Spec.Channel`, mirroring the behavior of `runUpgradeWithChannelVersion`. Image-only upgrades (no channel provided) are unaffected.

## Testing

- Added a unit test covering the case where both image and a differing channel are set, asserting the channel patch is applied before the image patch.
- `go test ./pkg/clusterversion/...` passes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Image-based upgrades now synchronize the cluster’s update channel with the requested channel before applying the desired image.
  - Ensures channel and image updates are applied successfully and in the correct order.

### DIFF
--- a/pkg/clusterversion/clusterversion_test.go
+++ b/pkg/clusterversion/clusterversion_test.go
@@ -381,6 +381,39 @@ var _ = Describe("ClusterVersion client and utils", func() {
 					Expect(hasCommenced).To(BeTrue())
 				})
 			})
+			Context("When the clusterversion channel differs from the upgradeconfig and image is set", func() {
+				It("Updates the channel before setting the desired image", func() {
+					clusterVersion := configv1.ClusterVersion{
+						Spec: configv1.ClusterVersionSpec{
+							Channel: "stable-4.20",
+							DesiredUpdate: &configv1.Update{
+								Version: "Some version",
+							},
+						},
+					}
+					upgradeConfig.Spec.Desired.Channel = "stable-4.21"
+					upgradeConfig.Spec.Desired.Image = "quay.io/test/test-image"
+					channelPatch := client.RawPatch(types.MergePatchType, []byte(fmt.Sprintf(`{"spec":{"channel":"%s"}}`, upgradeConfig.Spec.Desired.Channel)))
+					imagePatch := client.RawPatch(types.MergePatchType, []byte(fmt.Sprintf(`{"spec":{"desiredUpdate":{"image":"%s","version":null}}}`, upgradeConfig.Spec.Desired.Image)))
+					gomock.InOrder(
+						mockKubeClient.EXPECT().Get(gomock.Any(), gomock.Any(), gomock.Any()).SetArg(2, clusterVersion).Return(nil),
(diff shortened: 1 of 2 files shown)


simonkrenger:
Hello @pazpuwi I forwarded this PR to the SRE Operators team, they had some feedback regarding one of the findings above:

> I gave it a quick glance and looks good initially, but I feel the comment about injecting the values directly into the format-string is valid and would be good to be addressed so it can't inject strange data into the spec.

Could you review the finding and check if we can marshal a controlled payload with encoding/json at both sites? I see the other code is also doing this, but I feel it would be good to handle this as a start here

